PR #175, opened 2026-07-12
fix: harden RemoveOrphanedFilesScheduler against disposal race and DST clock skew

### LINKED ISSUE #171: fix: harden RemoveOrphanedFilesScheduler against disposal race and DST clock skew
## Summary

Two defects in `RemoveOrphanedFilesSchedulerService` can cause a spurious `ObjectDisposedException` on schedule-config changes, and can make the daily orphan-cleanup run early/late (including double-run or skip) across DST or other wall-clock adjustments.

@Kha-kis identified these and has a candidate fix ([ebb3f7e](https://github.com/Kha-kis/nzbdav/commit/ebb3f7e613c76cc4d48a5e69c464a9004eb2e422)); they want to participate in the repair.

## Defect 1: Dispose race on `_rescheduleCts`

The config handler exchanges `_rescheduleCts` and disposes the old source:

```csharp
var old = Interlocked.Exchange(ref _rescheduleCts, new CancellationTokenSource());
old.Cancel();
old.Dispose();
```

`ExecuteAsync` can be between its field read and `.Token` access when building linked token sources. `CancellationTokenSource.Token` throws `ObjectDisposedException` once disposed. That exception is not `OperationCanceledException`, so it hits the generic error path (logged + 10s backoff) instead of a clean reschedule.

**Fix direction:** cancel without disposing (cancelling is what wakes the loop; linked sources built in `ExecuteAsync` are still disposed by their `using`), and read the field once per iteration (e.g. `Volatile.Read`).

## Defect 2: Single relative `Task.Delay` vs wall clock

A single `Task.Delay(nextRun - now)` is computed against `DateTime.Now`. A DST shift or clock adjustment during the wait makes it fire an hour early or late, and can double-run or skip a day (e.g. fall-back early wake → task runs → loop still sees today's schedule ahead → runs again).

**Fix direction:** sleep in ≤30-minute slices and re-check the wall clock (`if (DateTime.Now < nextRun) continue`), logging the target only when it changes.

## References

- Code: [`backend/Services/RemoveOrphanedFilesSchedulerService.cs`](https://github.com/nzbdav/nzbdav/blob/main/backend/Services/RemoveOrphanedFilesSchedulerService.cs)
- Candidate fix: https://github.com/Kha-kis/nzbdav/commit/ebb3f7e613c76cc4d48a5e69c464a9004eb2e422

## Acceptance criteria

- [ ] Config changes to the orphan-schedule settings never throw `ObjectDisposedException` from `_rescheduleCts.Token`
- [ ] Long waits re-evaluate against local wall clock so DST/NTP mid-wait cannot early-fire / double-run the daily task
- [ ] Next-run log line is not spammed on slice wakes

Kha-kis:
Fixes #171.

Two issues in `RemoveOrphanedFilesSchedulerService`:

**Disposal race.** The config-changed handler does `Interlocked.Exchange` on `_rescheduleCts` and disposes the old source, while `ExecuteAsync` may be between reading the field and accessing `.Token`. `CancellationTokenSource.Token` throws `ObjectDisposedException` once disposed, which lands in the generic error path (10s backoff). Cancel without disposing (the old source becomes unreferenced and is GC'd), and read the field once per iteration via `Volatile.Read`.

**DST / clock skew.** A single `Task.Delay(nextRun - now)` computed off `DateTime.Now` fires early or late across a DST shift or clock adjustment. Sleep in <=30-minute slices and re-check the wall clock instead. A backward shift (DST fall-back or NTP step) re-exposes a run time already fired, so the last run is remembered and today's slot is treated as pending only when strictly later, keeping it to once per day. The "next run" log is emitted only when the target changes, and that state is reset when the schedule is disabled so re-enabling re-announces the run.

Verified manually by simulating the loop's scheduling decisions across a DST fall-back (fires once, not twice), normal multi-day operation (once per day), and a same-day disable/re-enable (re-arm is logged).

### DIFF
--- a/backend/Services/RemoveOrphanedFilesSchedulerService.cs
+++ b/backend/Services/RemoveOrphanedFilesSchedulerService.cs
@@ -29,9 +29,17 @@ public class RemoveOrphanedFilesSchedulerService : BackgroundService
 
             var old = Interlocked.Exchange(ref _rescheduleCts, new CancellationTokenSource());
             old.Cancel();
-            old.Dispose();
+            // Not disposed: ExecuteAsync may access .Token on this source after the swap, which
+            // throws ObjectDisposedException once disposed. Cancelling wakes the loop; the old
+            // source is then unreferenced and GC'd.
         };
     }
+
+    // Sleep in <=30m slices and re-check the wall clock, so a DST/NTP shift mid-wait can't fire
+    // the daily run early, twice, or not at all (a single Task.Delay(nextRun-now) would).
+    private static readonly TimeSpan MaxSleepSlice = TimeSpan.FromMinutes(30);
+    private DateTime? _lastLoggedNextRun;
+    private DateTime? _lastRun;
 
     protected override async Task ExecuteAsync(CancellationToken stoppingToken)
     {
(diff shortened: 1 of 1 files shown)


hoivikaj (approved):
LGTM

PR #176, opened 2026-07-12
fix(health): isolate arr failures in repair and skip empty root paths

### LINKED ISSUE #170: fix(health): Repair() aborts on one Arr failure and can throw on null root Path
## Summary

`HealthCheckService.Repair()` iterates configured Radarr/Sonarr instances to find which one owns an unhealthy linked file, then falls through to `File.Delete(symlinkOrStrmPath)` when none claims it.

Today there is **no per-instance error isolation**. An exception from `GetRootFolders()` or `RemoveAndSearch()` on one Arr aborts the entire loop (remaining instances are never tried). The outer `catch` then sets `RepairStatus=ActionNeeded` and leaves the file in place — so throwing failures are not silently destructive, but multi-Arr setups are fragile: a transient outage on an unrelated instance can block a successful repair on the instance that actually owns the file.

Worse, a naive “skip failing instance and continue” would be unsafe: an Arr we cannot reach may still own the file, and fall-through would then delete a link it still tracks.

Root folders with a **null `Path`** are also unhandled: `symlinkOrStrmPath.StartsWith(x.Path!)` throws `ArgumentNullException`, which likewise aborts the loop via the outer catch.

cc @Kha-kis — interested in participating on this.

## Current behavior

Relevant loop in [`backend/Services/HealthCheckService.cs`](backend/Services/HealthCheckService.cs):

```csharp
foreach (var arrClient in _configManager.GetArrConfig().GetArrClients())
{
    var rootFolders = await arrClient.GetRootFolders()...;
    if (!rootFolders.Any(x => symlinkOrStrmPath.StartsWith(x.Path!))) continue;

    if (await arrClient.RemoveAndSearch(symlinkOrStrmPath)...)
    {
        // Repaired — return
    }

    break; // fall through to File.Delete
}

await Task.Run(() => File.Delete(symlinkOrStrmPath))...; // Deleted
// outer catch → ActionNeeded, keep file, retry in 1 day
```

Contrast: [`ArrMonitoringService`](backend/Services/ArrMonitoringService.cs) already wraps each Arr host in its own try/catch.

## Expected behavior

1. Catch failures **per Arr instance** and keep trying the others.
2. Track that ownership could not be fully determined when any instance failed.
3. If any instance failed, record `RepairStatus=ActionNeeded` and **leave the file in place** instead of deleting it.
4. Skip root folders with a null (or empty) `Path` instead of throwing inside `StartsWith`.

## Related residual risk (optional follow-up)

When a root-folder prefix matches but `RemoveAndSearch` returns `false`, the code `break`s and deletes. That is a separate unsafe path (e.g. wrong Arr matched by overlapping roots). This issue focuses on throwing / unreachable-instance handling and null `Path`.

## Test plan (when fixed)

- [ ] One Arr throws on `GetRootFolders`; another owns the file → still repairs via the healthy instance (or ActionNeeded if ownership indeterminate).
- [ ] Unreachable Arr that may own the path → ActionNeeded, **no** `File.Delete`.
- [ ] Null/empty root-folder `Path` → skipped; other folders/instances still evaluated.
- [ ] All instances reachable and none claim the path → existing delete fall-through unchanged.

Kha-kis:
Fixes #170.

`Repair()` iterates configured arr instances to find which owns an unhealthy linked file, falling through to `File.Delete` when none claims it.

**Loop abort.** An exception from `GetRootFolders()` or `RemoveAndSearch()` on one instance aborted the whole loop, so a transient outage on an unrelated arr blocked repair on the instance that actually owns the file. Catch per instance and keep trying the others.

**Ownership resolution.** With per-instance isolation, deletion must only happen when ownership was actually resolved. The fix tracks two outcomes separately: whether an owning instance confirmed the file is an orphan, and whether any instance could not be reached or fully queried. It deletes only when an owner confirmed the orphan, or every instance was reached and none claimed the file. When ownership is indeterminate it records `ActionNeeded`, leaves the file in place, and defers the next check by a day so the item is not re-selected on every scan cycle while the instance stays down.

**Null root Path.** `StartsWith(x.Path!)` threw on a null Path. Root folders with a null or empty Path are skipped; when that is the only reason an instance did not match, it is treated as indeterminate rather than deleting a link the instance may still own.

Verified manually by driving `Repair()` against a reachable arr plus an unreachable one: an unreachable instance leaves the file in place and defers the next check; an owner that confirms the orphan still deletes even while another instance is down; and a genuine orphan (all instances reached, none owns it) is still deleted.

### DIFF
--- a/backend/Services/HealthCheckService.cs
+++ b/backend/Services/HealthCheckService.cs
@@ -1,5 +1,6 @@
 ﻿using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Hosting;
+using NzbWebDAV.Clients.RadarrSonarr.BaseModels;
 using NzbWebDAV.Clients.Usenet;
 using NzbWebDAV.Config;
 using NzbWebDAV.Database;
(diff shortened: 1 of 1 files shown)


hoivikaj (approved):
LGTM

PR #287, opened 2026-07-13
fix(auth): hmac-key password verification cache

hoivikaj:
## Summary
- Replace `PasswordUtil` plaintext `CacheKey(hash, password, salt)` with an HMAC-SHA256 hex key over those fields using a process-lifetime secret.
- Add 5-minute sliding expiration (match WebDAV credential cache); keep `SizeLimit = 5`.
- Extend `UtilityTests` to confirm verify + cache-hit correctness.

## Reasoning
Plaintext in the cache key table keeps secrets resident for the entry lifetime. HMAC keys preserve hit behavior for Rclone without `--use-cookies` while matching `ServiceCollectionAuthExtensions.VerifyPasswordWithCache`.

## Test plan
- [x] `dotnet test … --filter FullyQualifiedName~UtilityTests.Password`
- [ ] Wrong password still fails; correct password still succeeds on repeat verifies

Fixes #232

PR #288, opened 2026-07-13
fix(auth): close username-enumeration timing oracle

### LINKED ISSUE #230: Eliminate username-enumeration timing in authentication
## Source
Backend security & performance audit (`todo_backend.md` item 3), review date 2026-07-11.

## Severity
MEDIUM

## Where
- [AuthenticateController.cs](backend/Api/Controllers/Authenticate/AuthenticateController.cs#L13-L23)
- [ServiceCollectionAuthExtensions.cs](backend/Auth/ServiceCollectionAuthExtensions.cs#L63-L64)

## Problem
In `AuthenticateController`, when the account row is not found the expensive
`PasswordUtil.Verify` is skipped entirely, so "unknown user" responds measurably faster than
"known user, wrong password". Similarly, WebDAV basic auth compares the username with `==` and
short-circuits (`context.Username == user && VerifyPasswordWithCache(...)`), skipping hash
verification for wrong usernames. This lets an attacker confirm valid usernames by timing.

## Suggested fix (from audit)
**Non-breaking:** When the account/user does not match, still run `PasswordUtil.Verify`
against a static dummy PBKDF2 hash and discard the result, and compare usernames with
`FixedTimeEquals`. Response contracts stay identical.

## Suggested order
Items 2, 3, 7 (constant-time comparisons + dummy-hash) — mechanical, high confidence.

---

hoivikaj:
## Summary
- Always run `PasswordUtil.Verify` on UI login and WebDAV basic auth; when the account/user is missing, verify against a static dummy PBKDF2 hash and discard the result.
- Compare WebDAV usernames with `FixedTimeEquals`.
- Add `PasswordUtil.VerifyDummy` shared by both call sites.

## Reasoning
Skipped PBKDF2 is the dominant timing oracle for username enumeration. Response contracts stay identical (`Authenticated` bool / WebDAV 401). Residual DB-lookup timing is accepted and out of scope.

**Depends on:** #287 (HMAC cache keys in `PasswordUtil`). Retarget base to `main` after that merges.

## Test plan
- [x] `dotnet test … --filter FullyQualifiedName~UtilityTests.Password`
- [ ] Unknown username vs wrong password both return `Authenticated: false` / WebDAV 401
- [ ] Valid credentials still succeed

Fixes #230

### DIFF
--- a/backend/Api/Controllers/Authenticate/AuthenticateController.cs
+++ b/backend/Api/Controllers/Authenticate/AuthenticateController.cs
@@ -15,11 +15,22 @@ public class AuthenticateController(DavDatabaseClient dbClient) : BaseApiControl
             .Where(a => a.Type == request.Type && a.Username == request.Username)
             .FirstOrDefaultAsync().ConfigureAwait(false);
 
+        // Always run Verify (dummy hash when account is missing) so timing does not
+        // enumerate valid usernames by skipping the expensive PBKDF2 path.
+        var verified = account != null
+            ? PasswordUtil.Verify(account.PasswordHash, request.Password, account.RandomSalt)
+            : RunDummyVerify(request.Password);
+
         return new AuthenticateResponse()
         {
-            Authenticated = account != null
-                && PasswordUtil.Verify(account.PasswordHash, request.Password, account.RandomSalt)
+            Authenticated = account != null && verified
         };
+    }
+
+    private static bool RunDummyVerify(string password)
+    {
+        PasswordUtil.VerifyDummy(password);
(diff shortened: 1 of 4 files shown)


hoivikaj:
LGTM — reviewed: dummy-hash verify runs on both missing-account (UI login) and username-mismatch (WebDAV basic auth) paths, result always discarded so there is no bypass via the known dummy password; `FixedTimeEquals` username compare; wire contracts unchanged; cache-hit timing is symmetric between unknown-user and wrong-password repeats. CI green.

PR #289, opened 2026-07-13
fix(ui): safe-decode paths in proxy auth and compression

### LINKED ISSUE #217: Safe decodeURIComponent in proxy/auth/compression
## Source
Frontend security & performance audit (`todo_frontend.md` item 3), review date 2026-07-11.
Scope: `frontend/` (React Router 7 SSR + Express 5 proxy layer, `ws`).
`npm audit` (prod and dev): **0 known vulnerabilities** at audit time.

Companion audits: `todo_backend.md`, `todo_general.md`.

## Severity
LOW

## Where
[frontend/server/proxy-path.ts](frontend/server/proxy-path.ts#L13),
[frontend/app/auth/auth-middleware.server.ts](frontend/app/auth/auth-middleware.server.ts#L19),
[frontend/server.ts](frontend/server.ts#L31)

## Problem
A request path with malformed percent-encoding (e.g. `/%zz`, `/view%`) makes `decodeURIComponent` throw `URIError` inside the proxy gate, the auth middleware, and the compression filter. Express 5 converts these to 500s (error page + stack in logs) rather than a clean 4xx, and unauthenticated clients can generate them at will — noisy-log/error-page DoS, and a latent crash risk if any of these ever run outside Express's catcher.

## Suggested fix (from audit)
**Non-breaking:** Add a `safeDecode(path): string | null` helper (try/catch) used by all three call sites; on `null`, treat as "does not match any allowlist/public path" (falls through to auth redirect) or respond 400. Valid URLs behave exactly as before. Add a unit test next to [proxy-path.test.ts](frontend/server/proxy-path.test.ts).

## Suggested order
Item 3 + 4 (safe decode + allowlist tightening) — small, tested via `proxy-path.test.ts`. First in suggested attack order.

---
_Tracked from fable frontend audit (Jul 2026). Label: `fable-audit`._

hoivikaj:
## Summary
- Add shared `safeDecodePath` that returns `null` on malformed percent-encoding.
- Use it in proxy allowlist, auth middleware, and compression skip filter.
- On `null`: do not proxy; treat as non-public (auth redirect); do not skip compression.

## Reasoning
`decodeURIComponent` throws `URIError` on `/%zz` etc., which Express 5 turns into noisy 500s for unauthenticated clients. Fail closed rather than inventing a path.

## Test plan
- [x] Vitest: `proxy-path.test.ts`, `auth-middleware.server.test.ts`
- [ ] `/%zz` no longer produces a 500 from proxy/auth/compression gates

Fixes #217

### DIFF
--- a/frontend/app/auth/auth-middleware.server.test.ts
+++ b/frontend/app/auth/auth-middleware.server.test.ts
@@ -75,13 +75,14 @@ describe("authMiddleware", () => {
     expect(isAuthenticatedMock).not.toHaveBeenCalled();
   });
 
-  it("throws on malformed URI paths", async () => {
+  it("redirects malformed URI paths instead of throwing", async () => {
+    isAuthenticatedMock.mockResolvedValueOnce(false);
     const next = vi.fn();
     const res = mockRes();
 
-    await expect(
-      authMiddleware(mockReq("/%E0%A4%A"), res, next),
-    ).rejects.toThrow(URIError);
+    await authMiddleware(mockReq("/%E0%A4%A"), res, next);
+
+    expect(res.redirect).toHaveBeenCalledWith(302, "/login");
     expect(next).not.toHaveBeenCalled();
   });
 });
--- a/frontend/app/auth/auth-middleware.server.ts
+++ b/frontend/app/auth/auth-middleware.server.ts
@@ -1,5 +1,6 @@
 import type express from "express";
 import { isAuthenticated } from "~/auth/authentication.server";
+import { safeDecodePath } from "../../server/proxy-path";
 
 // Paths that do not require authentication. Every other path is protected.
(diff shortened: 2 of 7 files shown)


hoivikaj:
LGTM — reviewed against #217: `safeDecodePath` fails closed everywhere (not proxied / not public / compression default), `URIError` 500s eliminated, existing auth test updated from expect-throw to expect-redirect. Two fixes made during review: (1) added `server/proxy-path.ts` to `tsconfig.node.json`'s explicit include list — the original push broke frontend CI typecheck (TS6307); (2) the credential rate limiter's `skip` callback still called bare `decodeURIComponent`, the same throwing pattern this PR removes — now uses `safeDecodePath` (malformed path = not a credential path). Typecheck, build, and all 119 vitest tests pass; CI green.

PR #290, opened 2026-07-13
fix(ui): match proxy allowlist on path segment boundaries

### LINKED ISSUE #218: Tighten proxy allowlist (remove /p/, segment boundaries)
## Source
Frontend security & performance audit (`todo_frontend.md` item 4), review date 2026-07-11.
Scope: `frontend/` (React Router 7 SSR + Express 5 proxy layer, `ws`).
`npm audit` (prod and dev): **0 known vulnerabilities** at audit time.

Companion audits: `todo_backend.md`, `todo_general.md`.
Cross-ref: closed general-audit issue #182 tracked the dead `/p/` hygiene angle of this same finding.

## Severity
LOW

## Where
[frontend/server/proxy-path.ts](frontend/server/proxy-path.ts#L1-L19),
[frontend/server.ts](frontend/server.ts#L31-L44)

## Problem
Every entry in `BACKEND_PATH_PREFIXES` is an unauthenticated passthrough to the backend, so the list should be minimal and precise. (a) `/p/` matches no backend route (grep of `[Route(...)]` shows only `/api`, `/view`, `/nzbs`, `/content`, `/completed-symlinks`, `/.ids`, `/adapters/...`) — dead attack surface. (b) Bare-prefix `startsWith` means `/apifoo`, `/viewport.css`, `/contents-page` are all shipped to the backend instead of the frontend router, which both misroutes hypothetical future frontend pages and widens the unauthenticated surface.

## Suggested fix (from audit)
**Non-breaking:** Remove `/p/` (verify against git history that nothing external was ever documented to use it), and match on path-segment boundaries: `path === prefix || path.startsWith(prefix + "/")`. No currently-valid URL changes behavior. Also align `setApiKeyForAuthenticatedRequests` in [app.ts](frontend/server/app.ts#L40-L55) to use the same decoded path as `shouldProxyToBackend` (today one decodes and the other does not — currently fail-closed, but it is a latent inconsistency).

## Suggested order
Item 3 + 4 (safe decode + allowlist tightening) — small, tested via `proxy-path.test.ts`. First in suggested attack order.

---
_Tracked from fable frontend audit (Jul 2026). Label: `fable-audit`._

hoivikaj:
## Summary
- Match backend prefixes on segment boundaries (`path === prefix || path.startsWith(prefix + \"/\")`), normalizing trailing-slash prefixes like `/adapters/`.
- Align `setApiKeyForAuthenticatedRequests` with the same decoded `/api` check.
- `/p/` removal was already done in #201 — not repeated here.

## Reasoning
Bare `startsWith(\"/api\")` widens the unauthenticated proxy surface to `/apifoo` etc. Aligning API-key injection with decoded paths removes the latent encoded-`/api` inconsistency.

**Depends on:** #289 (safeDecode). Retarget base to `main` after that merges.

## Test plan
- [x] Vitest: `proxy-path.test.ts`, `inject-api-key.server.test.ts`
- [ ] `/apifoo`, `/viewport.css` are not proxied; `/api/foo` still is

Fixes #218

### DIFF
--- a/frontend/server/inject-api-key.server.test.ts
+++ b/frontend/server/inject-api-key.server.test.ts
@@ -35,6 +35,19 @@ describe("setApiKeyForAuthenticatedRequests", () => {
     expect(req.headers["x-api-key"]).toBeUndefined();
   });
 
+  it("ignores /apifoo bare-prefix false positives", async () => {
+    const req = mockReq({ path: "/apifoo" });
+    await setApiKeyForAuthenticatedRequests(req);
+    expect(isAuthenticatedMock).not.toHaveBeenCalled();
+  });
+
+  it("injects for decoded /api paths", async () => {
+    isAuthenticatedMock.mockResolvedValueOnce(true);
+    const req = mockReq({ path: "/%61pi/get-config" });
+    await setApiKeyForAuthenticatedRequests(req);
+    expect(req.headers["x-api-key"]).toBe("injected-key");
+  });
+
   it("leaves an existing API key alone", async () => {
     const req = mockReq({
       path: "/api/get-config",
--- a/frontend/server/inject-api-key.server.ts
+++ b/frontend/server/inject-api-key.server.ts
@@ -1,5 +1,6 @@
 import type express from "express";
 import { isAuthenticated } from "~/auth/authentication.server";
+import { isBackendApiPath } from "./proxy-path";
 
 /**
  * Inject the frontend→backend API key for authenticated UI sessions that proxy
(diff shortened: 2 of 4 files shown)


hoivikaj:
LGTM — reviewed against #218: segment-boundary matching handles the `/adapters/` trailing-slash prefix correctly (normalized so `/adapters` exact and `/adapters/x` match, `/adaptersfoo` doesn't); `/apifoo`, `/viewport.css`, `/contents-page` no longer proxied; PROPFIND/OPTIONS behavior preserved; compression skip list now shares the same matcher instead of a duplicated inline copy; API-key injection aligned to the decoded segment-bounded `/api` check with encoded-path test coverage. `/p/` removal correctly skipped (done in #201). Merged the #289 tsconfig/rate-limiter fixes forward; CI green. Merge after #289, then retarget base to `main`.

Issue #158, opened 2026-07-12
feat(db): configurable SAB history retention


hoivikaj:
## Summary
Long-lived installs accumulate SAB history rows with no age-based pruning. This tracking issue covers adopting the history-retention idea from [elfhosted/nzbdav](https://github.com/elfhosted/nzbdav) / [PR #199](https://github.com/nzbdav-dev/nzbdav/pull/199), reimplemented safely for the current cleanup architecture.

## Requirements
- Configurable retention days (default 90; `0` = disable)
- Env override: `DATABASE_HISTORY_RETENTION_DAYS`
- Config key: `database.history-retention-days`
- Must use `RemoveHistoryItemsAsync(..., deleteFiles: false)` so mounted WebDAV/DavItems are preserved
- Settings → Maintenance UI control
- Document that retention cleans SAB history rows, not media content

## Out of scope
- Deleting mounted media automatically
- Brotli-in-SQLite / auto_vacuum=FULL (superseded by BlobStore)

hoivikaj:
## Related

Complementary follow-up: #236 — server-side ceiling on SAB history page size per poll (default 10k), preserving full `noofslots`. Retention (#158) and the per-response cap address different failure modes.

PR #291, opened 2026-07-13
fix(sab): cap unbounded history responses

### LINKED ISSUE #236: Cap unbounded SAB history responses
## Source
Backend security & performance audit (`todo_backend.md` item 9), review date 2026-07-11.

## Severity
HIGH

## Where
[GetHistoryRequest.cs](backend/Api/SabControllers/GetHistory/GetHistoryRequest.cs#L10)
(`Limit { get; init; } = int.MaxValue`), used by
[GetHistoryController.cs](backend/Api/SabControllers/GetHistory/GetHistoryController.cs#L34-L35)

## Problem
When the *Arrs omit `limit`, or `history.ignore-sab-limit` is enabled, the query
materializes **every** history row (plus per-row slot mapping) into memory and serializes one
giant JSON document on every poll. Sonarr/Radarr poll frequently; with a multi-year history this
is repeated O(n) allocation, GC pressure, and SQLite scan work on the hottest API path.

## Suggested fix (from audit)
**Non-breaking:** Keep the "ignore the Arr's limit" semantics but introduce a high
server-side ceiling (e.g. configurable `history.max-page-size`, default 10 000 — far above what
Arrs import in one pass) and/or serialize with `IAsyncEnumerable` streaming so rows are not all
resident at once. Existing behavior is preserved for every realistic dataset while removing the
unbounded worst case.

## Suggested order
Item 9 (history ceiling) — biggest steady-state perf win for Arr-heavy installs.

---

hoivikaj:
## Summary
- Add `api.history-max-page-size` via `ConfigManager.GetHistoryMaxPageSize()` (default 10_000, clamp 1–100_000).
- Apply ceiling after Arr `limit` / ignore-sab-limit / frontend `pageSize` resolution in `GetHistoryRequest`.
- `TotalCount` / `noofslots` remain the full DB count; only `slots` is capped.

## Reasoning
Default `ignore-history-limit=true` plus `Limit=int.MaxValue` materializes full history on every Arr poll. A high ceiling preserves Arr import passes while removing the unbounded worst case. Retention (#158) shrinks the table over time but does not bound a single response.

## Test plan
- [x] `GetHistoryRequestTests`
- [ ] Arr ignore-limit path returns at most 10k slots with accurate total count
- [ ] Frontend `pageSize=100` unchanged

Fixes #236

### DIFF
--- a/backend/Api/SabControllers/GetHistory/GetHistoryRequest.cs
+++ b/backend/Api/SabControllers/GetHistory/GetHistoryRequest.cs
@@ -36,7 +36,8 @@ public class GetHistoryRequest
         //   * https://github.com/Sonarr/Sonarr/issues/5452
         //
         // Because of this, NzbDAV added a setting to ignore the `limit` value specified by the Arrs.
-        // When this setting is enabled, we always return all history items.
+        // When this setting is enabled, we skip the Arr limit and apply only the server-side ceiling
+        // (see GetHistoryMaxPageSize) so responses stay bounded.
         if (limitParam is not null && !configManager.IsIgnoreSabHistoryLimitEnabled())
         {
             var isValidLimit = int.TryParse(limitParam, out var limit);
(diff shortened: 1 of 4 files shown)


hoivikaj:
LGTM — reviewed against #236: ceiling applies after Arr-limit/ignore/pageSize resolution, `TotalCount` stays the full DB count, frontend `pageSize=100` unchanged, config default/clamp tested. One hardening added during review: switched `Math.Min` to `Math.Clamp(Limit, 0, ceiling)` because a client-supplied negative `pageSize`/`limit` would pass `Math.Min` and become a negative SQLite LIMIT — which is unbounded, exactly the bug this PR fixes. Test added.

PR #292, opened 2026-07-13
fix(api): compare profile tokens in constant time

### LINKED ISSUE #234: Compare profile/adapter tokens in constant time
## Source
Backend security & performance audit (`todo_backend.md` item 7), review date 2026-07-11.

## Severity
LOW

## Where
[ProfilePlayController.cs](backend/Api/Controllers/Profiles/ProfilePlayController.cs#L75)
(`Profiles.FirstOrDefault(x => x.Token == token)`) and the analogous lookups in the other
adapter controllers / `SearchProfileService`.

## Problem
These endpoints are deliberately reachable without the API key (auth *is* the URL
token), yet token matching is a short-circuiting ordinal comparison — same timing-leak class as
item 2, on an unauthenticated surface. Tokens are 96-bit CSPRNG values so practical risk is low,
but this is the cheapest surface to probe.

## Suggested fix (from audit)
**Non-breaking:** Compare candidates with `FixedTimeEquals` (iterate all profiles and
accumulate the match instead of `FirstOrDefault` with `==`).

## Suggested order
Items 2, 3, 7 (constant-time comparisons + dummy-hash) — mechanical, high confidence.

---

hoivikaj:
## Summary
- Add `ProfileConfig.FindByToken` that compares every profile with `FixedTimeEquals` and keeps the first match without exiting the loop early.
- Use it from `SearchProfileService.GetProfile` and `ProfilePlayController`.
- Use `FixedTimeEquals` for cache entry `ProfileToken` checks in NzbProxy and FailoverOrder.

## Reasoning
Unauthenticated adapter routes previously used ordinal `==` / `FirstOrDefault`, which can leak token prefix/position via timing. Profile lists are tiny, so a full scan is cheap. Hex tokens remain case-sensitive.

## Test plan
- [x] `ProfileConfigFindByTokenTests`
- [ ] Valid profile token still resolves; wrong token still 404

Fixes #234

### DIFF
--- a/backend/Api/Controllers/Profiles/Adapters/FailoverOrderController.cs
+++ b/backend/Api/Controllers/Profiles/Adapters/FailoverOrderController.cs
@@ -1,5 +1,6 @@
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
+using NzbWebDAV.Extensions;
 using NzbWebDAV.Services;
 using NzbWebDAV.Utils;
 using Serilog;
@@ -47,7 +48,7 @@ public class FailoverOrderController(
         foreach (var id in ids)
         {
             var entry = cache.Get(id);
-            if (entry is null || !string.Equals(entry.ProfileToken, token, StringComparison.Ordinal))
+            if (entry is null || !entry.ProfileToken.FixedTimeEquals(token))
                 continue;
 
             var c = entry.Primary;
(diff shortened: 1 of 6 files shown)


hoivikaj:
LGTM — reviewed against #234: `FindByToken` runs `FixedTimeEquals` against every profile before checking `match is null`, so no early exit leaks position; `ProfilePlayController`'s duplicate lookup now routes through it; NzbProxy/FailoverOrder cache-entry token compares also constant-time. WatchtowerService's `Token ==` intentionally untouched (config-vs-config, not attacker-controlled). Tests cover match/miss/duplicates. CI green.

PR #293, opened 2026-07-13
fix(webdav): handle malformed and unsatisfiable /view ranges

### LINKED ISSUE #233: Handle malformed and unsatisfiable Range headers
## Source
Backend security & performance audit (`todo_backend.md` item 6), review date 2026-07-11.

## Severity
LOW

## Where
[GetWebdavItemRequest.cs](backend/Api/Controllers/GetWebdavItem/GetWebdavItemRequest.cs#L46-L60)

## Problem
`long.Parse` on the range spec throws `FormatException`/`OverflowException` for
garbage like `Range: bytes=abc-` or a 30-digit number, surfacing as a logged 500 from the
generic handler. Malformed input at a system boundary should be rejected, not treated as a
server fault (it also pollutes error logs, and `/view` links are reachable with just a
download key).

## Suggested fix (from audit)
**Non-breaking:** Use `long.TryParse` and either ignore the invalid header (serve 200 full
content, matching RFC 7233 "ignore unparseable Range") or throw `BadHttpRequestException`.
Valid requests are unaffected.

## Suggested order
Rest as maintenance follow-ups.

---

hoivikaj:
## Summary
- Extract `TryParseRangeHeader` using `long.TryParse`; malformed/multi-range/overflow → leave range null and serve full 200 content.
- After `fileSize` is known, return 416 with `Content-Range: bytes */size` for unsatisfiable ranges (`start >= fileSize` or `start > end`).

## Reasoning
`long.Parse` turned garbage Range headers into logged 500s on a download-key surface. RFC 7233 says ignore unparseable Range (serve full content). Valid-but-unsatisfiable ranges get 416, matching WebDAV `GetAndHeadHandlerPatch`. NWebDav path malformed handling is out of scope (issue Where is `/view` only).

## Test plan
- [x] `GetWebdavItemRequestRangeTests`
- [ ] Manual: garbage Range → 200; `bytes=999999-` past EOF → 416; valid range → 206

Fixes #233

### DIFF
--- a/backend/Api/Controllers/GetWebdavItem/GetWebdavItemController.cs
+++ b/backend/Api/Controllers/GetWebdavItem/GetWebdavItemController.cs
@@ -79,8 +79,17 @@ public class GetWebdavItemController(
 
         if (rangeStart is not null)
         {
-            // compute
             var end = rangeEnd ?? (fileSize - 1);
+
+            // Syntactically valid but unsatisfiable → 416 (mirror WebDAV handler).
+            if (rangeStart.Value < 0 || rangeStart.Value >= fileSize || rangeStart.Value > end)
+            {
+                Response.Headers["Content-Range"] = $"bytes */{fileSize}";
+                Response.StatusCode = StatusCodes.Status416RangeNotSatisfiable;
+                await stream.DisposeAsync().ConfigureAwait(false);
+                return Stream.Null;
+            }
+
             var chunkSize = 1 + end - rangeStart.Value;
 
             // seek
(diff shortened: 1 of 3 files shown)


hoivikaj:
LGTM — reviewed against #233: malformed/multi-range/overflow Range headers now serve 200 full content (RFC 7233 ignore-unparseable) instead of 500 via bare `long.Parse`; valid-but-unsatisfiable ranges return 416 with `Content-Range: bytes */size`, mirroring `GetAndHeadHandlerPatch`. Edge cases verified: `bytes=-0` and suffix-on-empty-file both resolve to the 416 branch; `bytes=5-2` 416s; partially-parsed out-params are cleared on failure (follow-up commit). NWebDav WebDAV path correctly out of scope per the issue. CI green.

PR #295, opened 2026-07-13
fix(nntp): drain replaced clients before disposal

### LINKED ISSUE #235: Drain replaced NNTP clients before disposal
## Source
Backend security & performance audit (`todo_backend.md` item 8), review date 2026-07-11.

## Severity
LOW

## Where
[WrappingNntpClient.cs](backend/Clients/Usenet/WrappingNntpClient.cs#L87-L93)
(`ReplaceUnderlyingClient`), triggered from
[UsenetStreamingClient.cs](backend/Clients/Usenet/UsenetStreamingClient.cs#L21-L30)

## Problem
Saving usenet-provider settings swaps the client and immediately `Dispose()`s the
old one while streams may still be borrowing its connections, aborting in-flight playback/queue
downloads. Per AGENTS.md streaming invariants, completion callbacks must still fire — abrupt
pool disposal makes `NotRetrieved` failures spike at exactly the moment users tweak settings.

## Suggested fix (from audit)
**Non-breaking:** Drain instead of dispose: keep the old client alive until its connection
pool reports zero active connections (or a bounded grace period, e.g. 60s), then dispose it in
the background. New requests already go to the new client, so semantics only improve.

## Suggested order
Rest as maintenance follow-ups.

---

hoivikaj:
## Summary
- Swap `_usenetClient` immediately on provider config change; enqueue the old client for retirement.
- Background drain until `InFlightConnections == 0` (active borrows + pending selections) or 60s grace, then dispose.
- Bound retirement queue (max 4) for rapid successive saves; expose `INntpConnectionStats` through MultiProvider / Wrapping layers.

## Reasoning
Synchronous `Dispose()` on settings save aborts in-flight playback/queue borrows. Swap-first keeps new settings live; drain reduces `NotRetrieved` spikes. Forced dispose after grace may still fail in-flight work — goal is spike reduction, not eliminating forced teardown. Completion callbacks must still fire once per AGENTS.md invariants.

## Test plan
- [x] `WrappingNntpClientRetirementTests` (drain to zero; force after grace)
- [ ] Manual: change usenet providers during playback/queue download

Fixes #235

### DIFF
--- a/backend/Clients/Usenet/INntpConnectionStats.cs
+++ b/backend/Clients/Usenet/INntpConnectionStats.cs
@@ -1,0 +1,12 @@
+namespace NzbWebDAV.Clients.Usenet;
+
+/// <summary>
+/// Exposes in-flight connection counts so replaced clients can drain before dispose.
+/// </summary>
+public interface INntpConnectionStats
+{
+    /// <summary>
+    /// Borrowed connections plus pending provider selections still holding capacity.
+    /// </summary>
+    int InFlightConnections { get; }
+}
--- a/backend/Clients/Usenet/MultiConnectionNntpClient.cs
+++ b/backend/Clients/Usenet/MultiConnectionNntpClient.cs
@@ -78,6 +78,7 @@ public class MultiConnectionNntpClient(
     public int IdleConnections => connectionPool.IdleConnections;
     public int ActiveConnections => connectionPool.ActiveConnections;
     public int AvailableConnections => connectionPool.AvailableConnections;
+    public int InFlightConnections => ActiveConnections + PendingSelections;
 
     private int _pendingSelections;
     public int PendingSelections => Volatile.Read(ref _pendingSelections);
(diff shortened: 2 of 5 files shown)


hoivikaj:
LGTM — reviewed against #235: swap-first keeps new provider settings live immediately; drain polls `InFlightConnections` (active borrows + pending selections, correctly including `PendingSelections` which `ActiveConnections` alone misses); 60s grace then forced dispose; retirement queue bounded at 4 for rapid saves; `Dispose` flushes retiring clients. Streaming invariants preserved — `ConnectionPool.DisposeAsync` already lets borrowed connections finish and destroy-on-return, so callbacks still fire once. One fix during review: the test hook also kicked off the background drain loop, so two consumers raced on the retirement queue; the hook now drains inline only. Note `Serilog` usage inside `Clients/` (AGENTS.md pattern) is consistent with the rest of the directory. Tests cover drain-to-zero and forced-dispose-after-grace. CI green; the 11 local failures are the missing `rapidyenc` macOS native lib, unrelated.

PR #398, opened 2026-07-16
fix(ui): Overview live-stat row is visible against the page background

hoivikaj:
## Summary
- Add a subtle border and elevated `bg-base-200` fill to the Overview live-stat row so it no longer blends into the page background.

## Test plan
- [ ] Open Overview and confirm the top stats row (Active reads / Articles / throughput / Fetch errors) has a visible outline and elevated surface
- [ ] Confirm it still matches the look of other daisyUI stats rows (e.g. Watchdog) at desktop and narrow widths


hoivikaj:
Superseded by the Overview daisyUI migration PR (this contrast fix already landed on `main`).

PR #447, opened 2026-07-18
fix(webdav): small files no longer fail to play when players request a range past the end of the file

### LINKED ISSUE #448: [Bug]: /view returns invalid Content-Range when Range end is past EOF
### Before submitting

- [x] I am using NzbDav with legally obtained content.

### NzbDav version

current main / latest

### Installation method

Built from source

### Affected area

WebDAV or streaming

### What happened?

The `/view` endpoint does not clamp an HTTP Range end that runs past the end of the file. When `rangeStart` is in-bounds but `rangeEnd >= fileSize`, the handler skips the 416 check and emits an invalid `Content-Range` (end past size) plus an inflated `Content-Length`.

Native WebDAV already clamps with `Math.Min(range.End ?? length - 1, length - 1)` in `GetAndHeadHandlerPatch`. `/view` does not, leaving a gap after #233 (which covered malformed ranges and unsatisfiable starts, but not end-past-EOF clamping on `/view`).

Reported / fixed by @loambit in #447.

### Steps to reproduce

1. Serve a small file via `/view/...` (e.g. Jellyfin STRM or the web player).
2. Client sends a fixed probe/chunk range whose end is past EOF, e.g. `Range: bytes=0-1048575` when the file is only tens of KB.
3. Observe the `206` response: `Content-Range` end exceeds `fileSize`, and `Content-Length` is larger than the bytes that will actually be sent.

### Expected behavior

Clamp the range end to `fileSize - 1` so response headers stay valid (RFC 7233 “end past length → remainder of representation”), matching WebDAV behavior. Return a valid `206` with accurate `Content-Range` and `Content-Length`.

### User impact

Jellyfin STRM and web-player playback of small files can fail seek or stall when clients request oversized range ends — a common pattern for media players.

### Relevant logs or screenshots

_No response_

### Additional context

Fix: #447

loambit:
Fixes #448

Fixes an issue where small files can fail if the client requests a range past the end of the file (web player and Jellyfin STRM).

### DIFF
--- a/backend/Api/Controllers/GetWebdavItem/GetWebdavItemController.cs
+++ b/backend/Api/Controllers/GetWebdavItem/GetWebdavItemController.cs
@@ -104,7 +104,9 @@ public class GetWebdavItemController(
 
         if (rangeStart is not null)
         {
-            var end = rangeEnd ?? (fileSize - 1);
+            // clamp a range end that runs past the file to the last byte
+            // so the response headers stay valid.
+            var end = Math.Min(rangeEnd ?? (fileSize - 1), fileSize - 1);
 
             // Syntactically valid but unsatisfiable → 416 (mirror WebDAV handler).
             if (rangeStart.Value < 0 || rangeStart.Value >= fileSize || rangeStart.Value > end)


hoivikaj:
Superseding PR with your fix + regression tests: https://github.com/nzbdav/nzbdav/pull/451

PR #459, opened 2026-07-18
test(streams): fix flickering tests under parallel runs

loambit:
I noticed intermittent failures in the stream tests and traced it to `CollectingSink`. It appends to a plain `List<LogEvent>` with no synchronization, and since `Log.Logger` is process-global, every test class running in parallel emits into whichever sink a test currently has installed.


hoivikaj:
@loambit once again on top of it! When you make declarations of issues just so I am aware, is it safe to assume that you are mostly doing this testing from `main` or `latest` or `dev`?

loambit:
@hoivikaj this was on `main`

hoivikaj (approved):
LGTM

Issue #60, opened 2026-07-11
Pipeline STAT article-health checks


hoivikaj:
## Upstream reference

Upstream PR: nzbdav-dev/nzbdav#465 — https://github.com/nzbdav-dev/nzbdav/pull/465

## Audit status

**Still applies** (upstream PR is draft; depends on unreleased UsenetSharp `StatPipelinedAsync`)

## Problem

On-add and background article-health checks use sequential per-segment STAT calls. Pipelined STAT batches would speed queue processing, especially for large NZBs.

## Fork state

- `CheckAllSegmentsPipelinedAsync` exists in `backend/Clients/Usenet/NntpClient.cs`.
- `QueueItemProcessor` and `HealthCheckService` call `CheckAllSegmentsAsync` only (sequential path).
- Fork already has pipelining infrastructure (`usenet.pipelining.*` settings) but health checks are not wired to pipelined STAT.

## Recommendation

Wait for upstream UsenetSharp `StatPipelinedAsync` release (nzbdav-dev/UsenetSharp#5), then wire pipelined STAT through queue on-add and background health checks.

---
_Tracked from upstream open-PR audit (Jul 2026)._

hoivikaj:
## Proposed resolution

**Verdict:** Valid performance improvement — **infrastructure exists but not wired**. Pipelined STAT is implemented in `NntpClient.CheckAllSegmentsPipelinedAsync` / `StatsPipelinedAsync`, yet queue on-add and background health checks still call sequential `CheckAllSegmentsAsync`.

**Current fork touchpoints:**
- `backend/Clients/Usenet/NntpClient.cs` — `CheckAllSegmentsAsync`, `CheckAllSegmentsPipelinedAsync` (falls back to sequential on miss)
- `backend/Clients/Usenet/INntpClient.cs` — interface surface
- `backend/Queue/QueueItemProcessor.cs` (~L285) — on-add article existence uses `CheckAllSegmentsAsync`
- `backend/Services/HealthCheckService.cs` (~L169) — background health uses `CheckAllSegmentsAsync`
- `backend/Config/ConfigKeys.cs` / `ConfigManager.cs` — `usenet.pipelining.enabled`, `usenet.pipelining.depth`
- `docs/nntp-pipelining.md` — notes segment cache bypasses pipelined paths
- Upstream dependency: UsenetSharp `StatPipelinedAsync` (nzbdav-dev/UsenetSharp#5)

**Implementation steps:**
[...]

PR #470, opened 2026-07-18
feat(health): smooth out stat cliff bug, add aging function

loambit:
Last change from my own repo: after switching from v6 to v7 I noticed health checks were suspiciously fast, and realized there was a stat "cliff" at 8000 segments. v6 did 100% health checks on all files, which was thorough but slow, and v7 doesn't do enough, especially on larger files. To fix the cliff issue I applied a smooth curve so larger files fall off gracefully, as well as an aging function so recently posted items get more thorough health checks and old files get progressively lighter/faster ones. The default "Standard" option is roughly what the current check would do without the cliff, although I mostly run on "Deep" with a large library. I've been running this locally against `0.7.2X` for a while and figured it might be worth pushing up here, but also realize it might be a bit controversial. 

Coverage by depth setting (new releases without aging, note the cliff at 5.7GB in the current logic):

| segments |    size | current/0.7.2X | Standard | Enhanced | Deep | Complete |
|---------:|--------:|---------------:|---------:|---------:|-----:|---------:|
|    7,900 |  5.7 GB |           100% |     100% |     100% | 100% |     100% |
|    8,000 |  5.7 GB |            52% |     100% |     100% | 100% |     100% |
|   10,000 |  7.2 GB |            52% |      80% |      89% | 100% |     100% |
|   15,000 | 10.8 GB |            35% |      53% |      73% | 100% |     100% |
|   20,000 | 14.3 GB |            21% |      40% |      63% | 100% |     100% |
|   25,000 | 17.9 GB |            17% |      32% |      57% | 100% |     100% |
|   30,000 | 21.5 GB |            15% |      27% |      52% | 100% |     100% |
|   40,000 | 28.7 GB |            10% |      22% |      45% |  89% |     100% |
|   50,000 | 35.8 GB |             9% |      20% |      40% |  80% |     100% |
|   75,000 | 53.8 GB |             6% |      16% |      33% |  65% |     100% |
|  100,000 | 71.7 GB |             4% |      14% |      28% |  57% |     100% |

Coverage as a release ages (Standard depth):

| segments |    size |  new |   1y |  2y |  3y |  5y |  7y | 10y+ |
|---------:|--------:|-----:|-----:|----:|----:|----:|----:|-----:|
|    7,900 |  5.7 GB | 100% | 100% | 72% | 58% | 45% | 38% |  32% |
|    8,000 |  5.7 GB | 100% | 100% | 71% | 58% | 45% | 38% |  32% |
|   10,000 |  7.2 GB |  80% |  80% | 57% | 46% | 36% | 30% |  25% |
|   15,000 | 10.8 GB |  53% |  53% | 38% | 31% | 24% | 20% |  17% |
|   20,000 | 14.3 GB |  40% |  40% | 28% | 23% | 18% | 15% |  13% |
|   25,000 | 17.9 GB |  32% |  32% | 23% | 18% | 14% | 12% |  10% |
|   30,000 | 21.5 GB |  27% |  27% | 19% | 15% | 12% | 10% |   8% |
|   40,000 | 28.7 GB |  22% |  22% | 16% | 13% | 10% |  8% |   7% |
|   50,000 | 35.8 GB |  20% |  20% | 14% | 12% |  9% |  8% |   6% |
|   75,000 | 53.8 GB |  16% |  16% | 12% |  9% |  7% |  6% |   5% |
|  100,000 | 71.7 GB |  14% |  14% | 10% |  8% |  6% |  5% |   4% |

### DIFF
--- a/backend/Config/ConfigKeys.cs
+++ b/backend/Config/ConfigKeys.cs
@@ -63,6 +63,7 @@ public static class ConfigKeys
     public const string RepairEnable = "repair.enable";
     public const string RepairHealthcheckConcurrency = "repair.healthcheck-concurrency";
     public const string RepairHealthcheckDepth = "repair.healthcheck-depth";
+    public const string RepairHealthcheckAging = "repair.healthcheck-aging";
     public const string RepairAutoRemoveAfterFailures = "repair.auto-remove-after-failures";
     public const string RepairAutoRemoveUnlinkedOnly = "repair.auto-remove-unlinked-only";
     public const string ArrInstances = "arr.instances";
--- a/backend/Config/ConfigKeys.cs
+++ b/backend/Config/ConfigKeys.cs
@@ -62,6 +62,7 @@ public static class ConfigKeys
     public const string MediaLibraryDir = "media.library-dir";
     public const string RepairEnable = "repair.enable";
     public const string RepairHealthcheckConcurrency = "repair.healthcheck-concurrency";
+    public const string RepairHealthcheckDepth = "repair.healthcheck-depth";
     public const string RepairAutoRemoveAfterFailures = "repair.auto-remove-after-failures";
     public const string RepairAutoRemoveUnlinkedOnly = "repair.auto-remove-unlinked-only";
     public const string ArrInstances = "arr.instances";
--- a/backend/Config/ConfigManager.cs
+++ b/backend/Config/ConfigManager.cs
@@ -1125,7 +1125,7 @@ public class ConfigManager
     public double GetHealthCheckDepth()
     {
         var configured = StringUtil.EmptyToNull(GetConfigValue(ConfigKeys.RepairHealthcheckDepth));
-        return configured switch
+        return configured?.ToLowerInvariant() switch
         {
             "enhanced" => 1.0,
             "deep" => 2.0,
--- a/backend/Config/ConfigManager.cs
+++ b/backend/Config/ConfigManager.cs
@@ -14,6 +14,9 @@ public class ConfigManager
 {
     public static readonly string AppVersion = EnvironmentUtil.GetEnvironmentVariable("NZBDAV_VERSION") ?? "0.0.0";
 
+    // Sampling curve multiplier used by a background health check when none is configured.
+    public const double DefaultHealthCheckDepth = 0.5;
+
     private readonly Dictionary<string, string> _config = new();
     private readonly Dictionary<(string Name, Type Type), object?> _deserializedConfig = new();
     // Compiled exclude patterns are cached and rebuilt only when an exclude-related
@@ -1101,6 +1104,22 @@ public class ConfigManager
             ?? "50"
         );
         return Math.Clamp(configured, 1, Math.Max(1, poolSize));
+    }
+
+    /// <summary>
+    /// How much of each file a health check reads, as a multiplier on the sampling curve.
+    /// Returns 0 for "complete", which turns sampling off so every segment is checked.
+    /// </summary>
+    public double GetHealthCheckDepth()
+    {
+        var configured = StringUtil.EmptyToNull(GetConfigValue(ConfigKeys.RepairHealthcheckDepth));
+        return configured switch
+        {
+            "enhanced" => 1.0,
+            "deep" => 2.0,
+            "complete" => 0,
+            _ => DefaultHealthCheckDepth,
+        };
     }
 
     /// <summary>
--- a/backend/Config/ConfigManager.cs
+++ b/backend/Config/ConfigManager.cs
@@ -238,12 +238,17 @@ public class ConfigManager
                 case ConfigKeys.WardenHideDead:
                 case ConfigKeys.WardenBackboneScope:
                 case ConfigKeys.RepairEnable:
+                case ConfigKeys.RepairHealthcheckAging:
                 case ConfigKeys.RepairAutoRemoveUnlinkedOnly:
                 case ConfigKeys.RcloneRcEnabled:
                 case ConfigKeys.DbIsStartupVacuumEnabled:
                 case ConfigKeys.MaintenanceRemoveOrphanedScheduleEnabled:
                 case ConfigKeys.BackupScheduleEnabled:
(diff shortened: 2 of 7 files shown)


hoivikaj:
Before I mess with this component just yet, let me get https://github.com/nzbdav/nzbdav/issues/60 in. Not saying they collide or anything I had just already started some work on this (which updates some of our managed/forked deps).

hoivikaj:
@loambit 

**Note: this review was produced by a code review agent (run against the PR diff and current `main`).**

### Summary

The cliff diagnosis is correct: on `main`, `stride = count / 4000` with integer division gives 100% coverage up to 7,999 segments and drops to ~50% at exactly 8,000. The replacement is well built — `SampleTarget`'s `depth * sqrt(8000 * count)` curve is smooth, the carry-based remainder loop picks exactly `target` evenly spaced indices (no rounding step anywhere on the curve), head/tail coverage and the urgent-repair path are preserved, age is computed after `UpdateReleaseDate` backfills, and negative ages clamp safely. Test coverage is thorough. CI is green.

### Issues

1. **The "always checked in full below 8,000 segments" claim is contradicted by the aging behavior.** `AgeWeight` is applied *after* the `Math.Max(SampleFloor, …)` clamp, so a 7,900-segment file at 10 years gets `min(7900, 8000 × 0.316) ≈ 2529` segments (~32%) — the PR's aging table shows this honestly, but the `SampleFloor` code comment and the settings help text both claim a hard floor. Either apply the floor after aging (`Math.Max(SampleFloor, curve * AgeWeight(age))`) if a true floor is intended, or reword both. `SampleSegments_ChecksSmallFilesInFull` only runs without age, so it doesn't pin down which behavior is intended — please add a test that does.

2. **Aging compounds with the existing schedule backoff.** `ComputeNextHealthCheck` already doubles the check interval with age; this adds a second age-based reduction on top, so old items are checked both exponentially less often *and* up to ~3x less thoroughly per check. DMCA takedowns mostly hit young posts, but retention-driven expiry hits old ones — exactly what this de-prioritizes. Worth making the aging half separately toggleable, or at least calling it out in release notes. The cliff fix alone is uncontroversial and could ship independently.

3. **Default load increases for recent/large files.** "Standard" roughly *doubles* coverage in the 15k–30k segment range vs current (21%→40% at 20k), per the PR's own table. Fine if intentional, but it's a default-behavior change for all users, not just a cliff fix.

4. **No validation for the new config key.** `ValidateConfigItems` in `ConfigManager.cs` validates every other repair key, but `repair.healthcheck-depth` accepts any string via the config API and silently falls through to Standard. A "one of: standard/enhanced/deep/complete" case would match the file's fail-fast pattern.

### Nits

- Commit scope: repo conventions use `health` for health-check/repair work, not `repairs`.
- In `AgeWeight`, `if (age is not { } posted)` binds a `TimeSpan` to a variable named `posted`, which reads as a date; in `PerformHealthCheck` the same name *is* a date. Rename one.
- Depth `0` meaning "complete" (maximum coverage) is a surprising sentinel; a nullable double or enum would read better.

### Safety properties verified

Worst-case coverage for any file is bounded below at ~2,530 STATs; files under ~2,500 segments are always fully checked regardless of age; the first/last 100 segments and index `count-1` are always included.

loambit:
Addressing the review points:

1. Tweaked the code comments, settings text, and added a test case.
2. I changed the health check aging feature to be a toggle (defaulted off), although I think the retention argument from the review agent is incorrect as all of the articles for a post will disappear over the retention cliff at the same time, so even the most basic health check would catch it. 
3. This was the purpose of the PR, to increase the health check coverage percentage and make it configurable.
4. Added validation

Fixed the nits except the one about the enum which would require a bit of a refactor, so I decided against doing that, maybe in a followup.

hoivikaj:
**Note: this review was produced by a code review agent (re-review of the updated PR against prior feedback).**

@loambit 

### Verdict

Previous issues are largely addressed. Aging is now opt-in and default-off, depth is validated, comments/tests pin the floor-before-aging behavior, and the rename/`health` scope nits look good. Happy to defer the depth-0-as-complete enum. Backend/frontend CI are green.

Agree with the retention pushback: if a provider’s retention cliff removes a whole post at once, even a light sample should catch it. With aging default-off, that debate no longer blocks the default path.

### Remaining issue

1. **`RequireOneOf` is case-insensitive, but `GetHealthCheckDepth` is case-sensitive.**

   Validation accepts `"Deep"` / `"COMPLETE"` via `StringComparer.OrdinalIgnoreCase`, then the getter’s exact-match `switch` falls through to Standard:

   ```csharp
   // ValidateConfigItems
   if (!allowed.Contains(value, StringComparer.OrdinalIgnoreCase)) …

   // GetHealthCheckDepth
   return configured switch
   {
       "enhanced" => 1.0,
       "deep" => 2.0,
       "complete" => 0,
       _ => DefaultHealthCheckDepth, // "Deep" / "COMPLETE" land here
   };
   ```

   So a valid API write can silently under-check. Normalize in the getter (`ToLowerInvariant()` before the switch), or make `RequireOneOf` ordinal to match the stored values.

### Minor

- Depth help text still says “Files up to 8000 segments are checked in full” without noting that aging (when enabled) can reduce that. Code comments and the new test are correct; a short cross-reference in the depth blurb would close the loop.
- First commit is still `feat(repairs): …`; squash/merge with the updated `feat(health): …` title is fine.

### Prior review checklist

| # | Prior finding | Status |
|---|---|---|
| 1 | Floor claim vs aging | Addressed (comments + `SampleTarget_FloorAppliesBeforeAgingNotAfter`); depth UI text still slightly imprecise |
| 2 | Aging compounds with schedule backoff | Addressed (opt-in, default off) |
| 3 | Default load increase | Accepted as intentional |
| 4 | Missing depth validation | Addressed (`RequireOneOf` + aging bool); see case bug above |
| Nit | `repairs` scope | Mostly fixed (PR title + follow-up commit) |
| Nit | `posted` vs age | Fixed (`elapsed`) |
| Nit | Depth `0` sentinel | Deferred — fine |

Once the case mismatch is fixed, this looks mergeable from a review standpoint (still subject to the #60 sequencing note).

PR #472, opened 2026-07-19
feat(nntp): speed up health checks and import existence probes with pipelined STAT

### LINKED ISSUE #60: Pipeline STAT article-health checks

hoivikaj:
## Summary
- Bumps `NzbDav.UsenetSharp` to **3.1.0** and wires `StatPipelinedAsync` for article existence checks.
- When **NNTP pipelining** is enabled, queue import existence checks and background health checks use a primary-provider pipelined STAT sweep, then recheck only misses with the existing concurrent failover path.
- Classifies batch STAT replies like single `StatAsync` (223 exists, 430/451 miss, other codes throw) so a stale connection goodbye cannot look like a missing article.

Closes #60

## Test plan
- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release` (STAT-focused tests pass; one unrelated macOS rapidyenc native-lib env failure in prefetch budget tests)
- [ ] Large NZB import with ensure-existence + pipelining on/off: compare wall time
- [ ] Health check on 4k+ segment file: faster with pipelining on; same healthy/unhealthy outcome
- [ ] Multi-provider: article only on backup still succeeds via miss recheck failover

### DIFF
--- a/backend/Clients/Usenet/BaseNntpClient.cs
+++ b/backend/Clients/Usenet/BaseNntpClient.cs
@@ -1,4 +1,5 @@
-﻿using NzbWebDAV.Clients.Usenet.Models;
+﻿using System.Runtime.CompilerServices;
+using NzbWebDAV.Clients.Usenet.Models;
 using NzbWebDAV.Exceptions;
 using NzbWebDAV.Extensions;
 using UsenetSharp.Clients;
@@ -15,6 +16,13 @@ namespace NzbWebDAV.Clients.Usenet;
 /// </summary>
 public class BaseNntpClient : NntpClient
 {
+    /// <summary>
+    /// Sweep chunk size for <see cref="StatsPipelinedAsync"/>. UsenetSharp windows STAT
+    /// internally with a sliding MaxPipelineDepth; this bound is only for progress
+    /// reporting and early-stop-on-miss granularity (not BODY pipelining depth).
+    /// </summary>
+    internal const int StatPipelinedSweepChunkSize = 512;
+
     private readonly IUsenetClient _client;
 
     public BaseNntpClient() : this(new UsenetClient(new UsenetClientOptions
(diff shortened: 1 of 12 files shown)


hoivikaj:
@loambit you mind looking at this real quick, since you just had healthchecks on the brain

loambit:
Think this was already merged, but noticed the fallback catches `UsenetUnexpectedResponseException`, but a session that dies mid sweep throws UsenetSharp's `UsenetProtocolException`, which is a different hierarchy so it escapes the catch and fails the check instead of falling over to the concurrent path, probably want to broaden that. Also minor but the raw `progress` in that catch restarts at 1 so the bar jumps backwards, and the `StatPipelinedSweepChunkSize` comment still mentions early-stop-on-miss which doesn't happen now that it drains the whole enumerable.

PR #476, opened 2026-07-19
fix(nntp): health checks no longer fail when a pipelined STAT session dies mid-sweep

hoivikaj:
## Summary
- When a pipelined STAT health/import sweep hits a mid-session failure (`UsenetProtocolException`, transport errors, etc.), fall back to the concurrent STAT path instead of failing the check.
- Keep the health-check progress bar from jumping backwards during that full fallback.
- Correct the `StatPipelinedSweepChunkSize` comment (no early-stop-on-miss).

Follow-up to review feedback on #472.

## Test plan
- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --filter CheckAllSegmentsPipelined`
- [ ] Health check with pipelining on: kill/idle a connection mid-sweep and confirm it completes via concurrent failover (same healthy/unhealthy outcome)
- [ ] Progress bar does not jump backwards when a mid-sweep failover occurs

### DIFF
--- a/backend/Clients/Usenet/BaseNntpClient.cs
+++ b/backend/Clients/Usenet/BaseNntpClient.cs
@@ -19,7 +19,8 @@ public class BaseNntpClient : NntpClient
     /// <summary>
     /// Sweep chunk size for <see cref="StatsPipelinedAsync"/>. UsenetSharp windows STAT
     /// internally with a sliding MaxPipelineDepth; this bound is only for progress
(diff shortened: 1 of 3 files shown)


hoivikaj:
fyi @loambit

loambit (approved):
The fallback swallows the exception without logging it, might be nice to have a debug log there to catch a provider who handles pipelining badly, otherwise lgtm

loambit (approved):

PR #482, opened 2026-07-19
fix(nntp): stop providers getting stuck in the probing state

loambit:
Bumped into this bug where a provider gets stuck in "Probing" state because a STAT failure trips the breaker but a STAT success never resets it. This makes the failure side match the success path, which already skips stat, head, and date.

update: Expanded so a successful stat, head, or date now closes the breaker when it's already latched, so a provider that sees little body traffic can recover on its own. Left get-connection alone as suggested, it stops being a trap now that a stat can close what it latched.

Screenshot 2026-07-19 110131

### DIFF
--- a/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
+++ b/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
@@ -74,6 +74,13 @@ public class ProviderCircuitBreaker
         }
     }
 
+    /// <summary>
+    /// True once a trip has latched, whether still cooling down or waiting on a probe.
+    /// Unlike <see cref="IsTripped"/> this claims no probe slot, so callers can read it
+    /// without altering state.
+    /// </summary>
+    public bool IsLatched => Volatile.Read(ref _trippedUntilMs) != 0;
+
     /// <summary>TickCount64 deadline while latched open; 0 when not tripped. For tests.</summary>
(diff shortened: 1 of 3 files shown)


hoivikaj:
@loambit  Comments from a code-review Agent:

Two comments:

The get-connection path can still reproduce this. The connection-acquisition failure handler in RunWithConnection records get-connection-{exception} failures regardless of which command triggered it. If a provider refuses connections while only STAT health checks are running, the breaker still trips from STAT-driven traffic, and once the provider recovers, the same stuck-Probing state recurs since nothing STAT-shaped can close the breaker. Connect/auth failures are arguably command-agnostic health signals worth counting, so I'm fine leaving it as is — but the underlying "nothing ever resets it" gap remains for providers that rarely see BODY traffic (backup providers especially). Might be worth a follow-up issue: e.g. allow any successful command to close the breaker only when a half-open probe is in flight, which wouldn't reintroduce the STAT-ok/BODY-fail masking the success-path comment warns about.

Nit: the new guard is an allow-list (name is "BODY" or "ARTICLE") while the success path below uses a deny-list (name is "STAT" or "HEAD" or "DATE"). Equivalent today, and allow-list is the safer default for future commands — just worth keeping the two from drifting.

loambit:
makes sense, I'll expand the fix so we're consistent everywhere.

PR #490, opened 2026-07-19
fix(queue): block arr thumbnails by default

loambit:
Adds .jpg to the blocklist, so `thumbnail.jpg` stops showing up in the health check queue.


hoivikaj:
@loambit  This is a good short term patch, I think the end result is still covered by this: https://github.com/nzbdav/nzbdav/issues/257 -- I just wnated to make sure you were aware that issue existed in case that alters any plans.

loambit:
Yup, I saw that one and agree it's 100% the correct long term approach!

PR #493, opened 2026-07-19
fix(nntp): only log provider recovery when the circuit actually opened

loambit:
My logs had seven "Provider X recovered - circuit breaker reset" lines in three minutes with no matching trip. RecordSuccess was logging whenever there was anything in the failure window to clear, so a single transient failure that never came close to tripping still announced a recovery. Dropped the window check so it only fires when the circuit actually opened, which is the same condition NotifyTransition below it already uses. Also added a shared collection for the two test classes that swap the global Log.Logger, since the suite runs collections in parallel and they would otherwise race.

### DIFF
--- a/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
+++ b/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
@@ -107,8 +107,11 @@ public class ProviderCircuitBreaker
     {
         lock (_lock)
         {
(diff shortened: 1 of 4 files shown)


hoivikaj:
@loambit Sorry github actions being down has sorta ruined my day. Are you in discord?

loambit:
No problem, this isn't critical. I'll pop in after the kids go to bed.

PR #525, opened 2026-07-21
feat: import altmount configuration

jackster1209:
## Summary

- Adds a guided **Altmount Migration** workflow under Settings with six steps: connect, map categories, scan, review, run, and optionally update library symlinks.
- Reads Altmount v2 metadata and `.nzbz` stores, reconstructs NZBs—including encryption and password metadata—and submits selected releases through NzbDAV's existing SAB queue and processing pipeline.
- Scans and classifies releases before submission, showing Green/Amber/Red verdicts, estimated fetch cost, unmapped categories, corrupt or missing stores, naming conflicts, and collisions with the live NzbDAV queue.
- Detects releases that are already present, avoids duplicate submissions, and records completed release/file provenance for future scans and symlink matching.
- Adds pause, resume, cancellation, submission reconciliation, migration-history cleanup, persisted preferences, and restart-safe background processing.
- Adds an optional dry-run symlink migration that can repoint Sonarr/Radarr/Plex library links from Altmount to their migrated NzbDAV files.
- Adds authenticated migration API endpoints and 143 focused backend tests covering metadata parsing, triage, collision detection, NZB reconstruction, submission reconciliation, provenance, and symlink planning/apply/restore behavior.

## Safety and compatibility

- Supports Altmount's shared-store format, including v2 and v3 metadata with `store_ref` references and zstd-compressed `.nzbz` stores.
- Legacy v1 metadata without a shared store reference is identified during review and excluded because it cannot be reconstructed losslessly.
- Scanning is read-only and does not make Usenet requests. Actual imports are submitted at low priority through NzbDAV's normal queue, with configurable worker and queue-depth limits.
- Migration state is stored in a separate, lazily created `${CONFIG_PATH}/usenet-migration.db`; the existing NzbDAV database schema and content are not modified by the migration database.
- Changing category mappings or relevant global naming settings requires a fresh scan before submission.
- Symlink migration is optional and begins with a reviewable dry-run. Only positively matched Altmount symlinks are eligible for rewriting.
- Currently only symlink migration is supported - strm migration may be added in a future update.
- Before applying a symlink plan, the wizard creates a restore archive. It leaves real files, unrelated links, already migrated links, and unresolved/orphaned links untouched, and refuses to rewrite links whose targets changed after planning.
- Resetting the wizard preserves migrated NzbDAV content and completed provenance unless the user explicitly chooses to forget those migration records.

## Test plan

- [x] `cd frontend && npm run typecheck`
- [x] `cd frontend && npm run build`
- [x] `cd frontend && npm test -- --run` — 214 passed
- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --filter FullyQualifiedName~UsenetMigration` — 143 passed
- [ ] Full backend test suite passes in GitHub Actions
- [x] Run connect → category mapping → scan → review against a representative Altmount v2 library
- [x] Confirm unmapped categories and blocking queue/name collisions prevent migration until resolved
- [x] Import selected Green and Amber releases and confirm completion is reconciled against NzbDAV queue/history
- [x] Restart NzbDAV during an active migration and confirm the run resumes without duplicate submissions
- [x] Plan symlink changes against a mixed library containing Altmount links, NzbDAV links, unrelated links, orphaned links, and real files
- [x] Apply a reviewed symlink plan, verify a restore archive is created first, and confirm restoring the archive returns links to their original targets


hoivikaj (changes requested):
Hello — I'm a **code review agent** working on behalf of a maintainer. I reviewed the full diff for this PR (~14.3k lines across the migration DB, import engine, symlink workflow, API surface, UI wizard, and the focused test suite). Overall this is an impressive piece of work: the isolated `usenet-migration.db`, read-only scan/triage model, provenance for already-migrated detection, restart-oriented runner, and backup-first symlink apply/restore are the right architectural shape for a high-stakes migration tool. The 143 backend tests in particular give a lot of confidence on the parsing/triage/symlink happy paths.

That said, I am **requesting changes** before merge. The issues below are correctness and safety problems in lifecycle / concurrency / path handling — not style nits. Several of them can produce duplicate SAB submissions, untracked in-flight work, or unexpected filesystem mutation if a user cancels, resets, or the process restarts at the wrong moment.

### Blocking findings

#### 1. Critical — `reset` is accepted during active work

`POST /api/altmount-migration/reset` has no session-status gate, unlike `migration-data/forget`, which correctly rejects `scanning | running | linking | applying`.

`UsenetMigrationStore.ResetAsync` deletes submissions, releases, the symlink plan, category map, and session state while `UsenetMigrationRunner` may still be mid-`SubmitBatchAsync` or mid-`SymlinkRewriter.ApplyAsync`.

**Why it matters:** In-flight NzbDAV queue jobs become untracked; a later scan can resubmit the same sources; symlink apply can finish after its plan rows were erased, leaving the UI claiming a clean wizard while filesystem work continues or has already partially applied.

**Ask:** Reject reset in the same active states as forget. Prefer requiring cancel → terminal state first. Disable/hide Reset in the UI (`ResetFooter`) while the session is active.

#### 2. High — Pause/cancel does not stop an in-flight submit batch

`CancelRunAsync` / pause only flip `session.Status`. `SubmissionWorkerPool.SubmitBatchAsync` snapshots all `pending` rows up front and continues calling `AddFileAsync` without re-reading session status. The `CancellationToken` it receives is the host stopping token, not a user-cancel CTS.

**Why it matters:** A user who cancels or pauses during a deep batch can still get many additional releases enqueued (up to `MaxQueueDepth`) after they believe the run stopped.

**Ask:** Introduce an operation epoch / CTS owned by the runner, and before each external `AddFileAsync` either honor that token or atomically verify the session is still `running`.

#### 3. High — Crash between `AddFileAsync` success and persistence can duplicate-submit and evict

In `SubmitBatchAsync`, NzoIds and `submitted` state are mutated in memory and `SaveChangesAsync` runs **once after the whole batch**. A process crash mid-batch loses those NzoIds; rows remain `pending`; restart resubmits. SAB `AddFileController` calls `ReplaceExistingQueueItemIfNeededAsync`, so the retry can **evict** the first import that was already progressing.

**Why it matters:** “Restart-safe” is a stated goal of this PR, but the submit path is not crash-safe across the external queue mutation boundary.

**Ask:** Persist a per-release `submitting` claim (and NzoId as soon as available) around each `AddFileAsync`. On recovery, locate the deterministic queue item before retrying; never blindly resubmit an ambiguous post-submit state.

#### 4. High — Shell injection via `LibraryRoot` through `SymlinkAndStrmUtil`

Symlink planning walks the library via `SymlinkAndStrmUtil.GetAllSymlinksAndStrms`, which builds:

```text
sh -c "cd '<path>' && find ..."
```

Only single quotes are escaped. A directory name containing `"` can break out of the outer double-quoted `-c` argument. This util already exists on main, but this PR newly wires user-supplied `libraryRoot` from `POST .../symlinks/plan` into it. API-key access is required, but that still means any authenticated admin/API-key holder can turn a migration setting into command execution as the backend user.

**Ask:** Do not interpolate paths into `sh -c`. Prefer `ProcessStartInfo.ArgumentList` with a fixed script and a positional root, or replace the shell walk with managed enumeration. As a stopgap, reject roots containing `"`, newlines, or other shell metacharacters.

#### 5. High — Failed UI mutations still advance the wizard

In `use-altmount-migration.ts`, `withBusy` catches errors, sets `error`, and **resolves successfully**. Call sites such as `saveCategories(...).then(onDone)` and `startRun().then(onRun)` therefore advance steps even when the API rejected the mutation. The Run step can then show “Migration running” copy while backend status is still `scanned`.

**Ask:** Re-throw after recording the error, or return a success boolean and only transition on success.

---

### Non-blocking but should fix soon

1. **Medium — Linux path drift compare is case-insensitive.** `SymlinkRewriter.PathsEqual` uses `OrdinalIgnoreCase` on every OS. On Linux, `/mnt/Altmount/x` and `/mnt/altmount/x` are distinct; a case-only target change can pass the drift guard and be overwritten. Prefer the OS-sensitive comparer already used in `SymlinkRestoreService`.

2. **Medium — Include/category edits during a run.** `PUT .../releases/include` is allowed while `running`. Clearing `Included` does not remove an already-materialized pending submission, so excluded releases can still import. Lock review mutations outside `scanned` / non-active states, or make include changes transactional with submission rows.

3. **Medium — Stale polling can overwrite fresh mutation results.** `refresh()` has no request generation / abort. An older poll can race a pause/resume/start/apply response and briefly re-enable wrong controls or show a stale step.

4. **Medium — Table load failures look like empty data.** Review/symlink table loaders that catch and render “No matches” on API failure hide safety-critical review failures. Surface the error and retain last-good data.

5. **Medium — Symlink restore/apply containment is lexical only.** `GetFullPath` checks without resolving/rejecting symlinked parents leave TOCTOU / parent-symlink escape residual risk for restore/apply. Worth hardening if this remains an admin filesystem tool.

6. **Low — Process / docs polish.** PR title should use conventional / user-facing form (e.g. `feat(ui): guided Altmount migration wizard`). Confirm that `feat(db)!` on an isolated lazy DB is an intentional breaking bump (primary DAV schema is untouched). Consider docs + a `since` pill for this Settings flow. Frontend tests for mutation failure / active reset blocking would help; backend tests for cancel-mid-batch and crash-after-AddFile gaps are currently missing.

---

### What looks solid (for balance)

- Dedicated SQLite under `${CONFIG_PATH}/usenet-migration.db`; primary DAV schema not modified by migration schema work; lazy create + hosted runner resume model.
- Scan is metadata-only; imports go through the normal SAB pipeline with queue-depth gating.
- Triage (Green/Amber/Red), collisions, already-migrated provenance, and reconciler treating vanished queue/history as terminal (anti-resubmit loop).
- Symlink apply: backup-first, target drift checks, confirm-gated apply/restore/forget.
- Consistent `GuardedAsync` API-key validation before DB init; restore archive filename constrained to the configured backup directory.
- Large, focused backend coverage for parsers, triage, reconciler, and symlink plan/apply/restore.

### Residual note

Even after the blockers above, an API-key holder can point metadata / library / backup paths at any process-visible filesystem location. That is consistent with other admin path settings in NzbDAV, but user docs should be explicit that symlink apply rewrites library links and should only be pointed at trusted roots.

Happy to re-review once the blocking items are addressed — especially reset gating, cancel/pause interruptibility, per-submit durability, the shell path, and the `withBusy` step-advance bug. Thanks again for the careful design and the thorough test investment on the core migration paths.

jackster1209:
I addressed the concerns from top to bottom, and the commit order reflects that - each focusing on a single presented issue at a time.

hoivikaj (changes requested):
Hello — I'm a **code review agent** working on behalf of a maintainer. I re-reviewed tip `4ecd6df1`. The previous findings were addressed well, including durable submission claims, batch interruption, shell-safe symlink scanning, UI transition gating, and stale-poll protection. Thank you for the focused fix series.

I am still requesting changes for these remaining safety and lifecycle issues:

### 1. High — Cancel becomes terminal before the current submission finishes

`StopRun(cancel: true)` stores `cancelled` and returns while an already-started `AddFileAsync` is deliberately allowed to continue. Reset and Forget then become available because `cancelled` is not active.

If the user resets before that call finishes, scan/submission rows are deleted; afterward `UpdateSubmissionAsync` can recreate a lone `submitted` row because it upserts missing submissions. The queue job is then outside a coherent wizard session. Without Reset, work completing under `cancelled` is no longer reconciled into provenance.

**Please add a `cancelling`/draining state, keep Reset and Forget blocked until the active queue boundary is quiescent, and recover/reconcile the final claim before publishing `cancelled`.** Add a test that cancels during a blocked `AddFileAsync` and immediately attempts Reset.

### 2. High — Symlink replacement can delete a real file during a leaf race

`CreateOrReplaceSymlink` records that the leaf is a symlink, re-checks only its parent chain, then deletes based on that stale result. If another process replaces the symlink with a regular file between those steps, `File.Delete` removes the real file. This breaks the documented “never delete real files” guarantee for both apply and restore.

**Re-read the leaf immediately before deletion and refuse unless it is still the expected symlink.** Please add fault-injection coverage for a symlink-to-file swap.

### 3. High — Connect and Scan can clobber active state

`POST /connect` always writes `Status = "connected"` with no active-work gate. `CanStartScan` blocks only `running|paused`, so Scan remains allowed during `linking|applying`. These transitions can stop reconciliation, overwrite a pending scan with `linked`, or clear migration rows while Step 6 is using them.

**Define explicit legal source states and claim transitions atomically. Reject Connect and Scan during every active operation, mirror that in the UI, and test the controller gates.**

### 4. High — Step 6 operations are not exclusive

Restore runs in the HTTP request while status remains `linked`. Apply or Rebuild Plan can therefore start concurrently and mutate/read the same library and plan rows. A client disconnect can also cancel restore after filesystem changes but before its DB updates are saved.

**Serialize plan/apply/restore behind a durable operation state such as `restoring`, or a shared single-flight gate. Move restore into the background runner if possible and disable all conflicting controls while any Step 6 operation is active.**

### Additional items

- **Scan Cancel is only best-effort:** the old scan can still persist results and set `scanned` after the UI reports cancellation.
- **Symlinked Library Root mismatch:** `find -H` plans through one, but apply/restore reject it. Reject it during planning or support it consistently.
- **`SubmitWorkers` has no effect:** it is persisted, shown, and documented, but submission remains sequential.
- **Step 6 failures can trap the wizard:** persistent plan/apply errors leave an active status while Reset remains blocked.
- **Known failures still emit stack dumps:** replace expected `Log.Warning(e, ...)` paths with human-friendly warnings per `AGENTS.md`.
- The guide H1 is missing `{ .nzbdav-since }`, and the PR title should use the user-facing conventional format.

Please add focused tests for the four High findings. I could not see PR checks reported for the head branch, so the full frontend and backend suites should also be green before merge.

The isolated migration database, metadata-only scan, provenance model, backup-first apply, API-key gating, and focused test coverage remain strong. Happy to re-review after the lifecycle and filesystem-safety blockers are addressed.

backend/Utils/SymlinkAndStrmUtil.cs
```diff
@@ -66,26 +39,25 @@ private static IEnumerable<ISymlinkOrStrmInfo> GetAllSymlinksAndStrmsLinux(strin
...
-            else
+
+            var target = new FileInfo(fullPath).LinkTarget;
+            if (target is not null)
             {
-                yield return new SymlinkInfo()
+                yield return new SymlinkInfo
                 {
-                    SymlinkPath = Path.GetFullPath(filePath, directoryPath),
+                    SymlinkPath = fullPath,
```
loambit: FYI this can delete library items for anyone with orphan cleanup enabled, because the library walk that got rewritten here is the same one that task uses to decide what is still linked. 

Just a suggestion, but since this import process only ever runs once, I'm wondering whether it would sit better as a standalone tool rather than woven into the main codebase.

@jackster1209 @hoivikaj

jackster1209: [Commit: isolate symlink library traversal](https://github.com/nzbdav/nzbdav/pull/525/commits/ded9bdb424563ef07353725debd35814b8f5a61b)

This should be sufficient to address your concern, which was definitely valid. Thanks for helping review this!

loambit: Hmm, the problem moved from `SymlinkAndStrmUtil` to `MigrationSymlinkUtil` along with the code. A symlink whose target does not resolve still gets dropped instead of planned, so it quietly stays pointing at Altmount while the migration reports complete. The comment above that loop says a failed scan should never hand back a partial plan, and that is what happens here when find succeeds.

jackster1209: I apologize, I misunderstood your first comment which led me to taking the wrong approach. The latest commit should resolve the issue you raised here.

jackster1209:
I've changed the base branch for the merge to "experiment-altmount-mig" as you requested. My PR branch is currently in sync with the upstream Main branch plus the 5 commits from my PR.

hoivikaj:
@jackster1209 thanks for the contrib! pushing into a working branch so I can auto bump it as main updates while we continue the test/review process. Will work on getting it into a dev tag shortly.

hoivikaj:
Fixed the backend CI failure in `e704a0f6`.

The restore lifecycle itself reached filesystem work, but the test then synchronously blocked the available xUnit worker in `BlockingSymlinkOps`. Under CI load, the test's `Task.Run` waiter could not run until that fake timed out after 30 seconds; restore then correctly returned to `linked`, producing the misleading `expected restoring / actual linked` failure.

The test now pauses asynchronously after restore has released its SQLite context and before filesystem work begins. This keeps the lifecycle assertions schedulable while preserving the original coverage: restore remains durably `restoring`, plan/apply are rejected, client disconnect does not cancel restore, and completion returns to `linked`.

Verification:
- Failed focused test: passed in 838 ms
- Full backend suite: 1,915 passed, 20 skipped, 0 failed

PR #529, opened 2026-07-21
fix(nntp): hard connect timeout so black-holed providers cannot hang callers

funkypenguin:
A provider whose hostname resolves but whose TCP layer black-holes (no SYN/ACK, no RST — a misconfigured firewall, or a DNS round-robin pointing at a dead host) hangs the connecting task for the OS SYN-retry window (~2 minutes), pinning a threadpool slot per attempt. The circuit breaker never trips for these providers, because `RecordFailure` only fires after the connect returns — so the worst kind of provider failure is the one the breaker can't see.

This wraps `ConnectAsync` in a linked cancellation with a hard cap (default 15s, tunable via `NNTP_CONNECT_TIMEOUT_SECONDS`) and converts the local timeout to `CouldNotConnectToUsenetException`, so the breaker counts it as a real failure rather than a caller cancel.

We run a large fleet of instances and this failure mode bites at scale — a single black-holed provider can quietly pin dozens of threadpool slots while every health indicator stays green.

### DIFF
--- a/backend/Clients/Usenet/BaseNntpClient.cs
+++ b/backend/Clients/Usenet/BaseNntpClient.cs
@@ -42,11 +42,33 @@ public class BaseNntpClient : NntpClient
         _client = client ?? throw new ArgumentNullException(nameof(client));
     }
 
+    // Hard cap on TCP+TLS+welcome-banner time per provider. Without this, a
+    // provider whose hostname resolves but whose TCP layer black-holes (no
+    // SYN/ACK, no RST — e.g. a misconfigured firewall or a DNS round-robin
+    // pointing at a dead host) hangs the calling task for the OS SYN-retry
+    // window (~2 minutes). That pins a threadpool slot per attempt and the
+    // provider circuit breaker never trips, because RecordFailure only fires
+    // after the connect returns.
+    private static readonly TimeSpan ConnectTimeout =
+        int.TryParse(Environment.GetEnvironmentVariable("NNTP_CONNECT_TIMEOUT_SECONDS"), out var v) && v > 0
+            ? TimeSpan.FromSeconds(v)
+            : TimeSpan.FromSeconds(15);
+
     public override async Task ConnectAsync(string host, int port, bool useSsl, CancellationToken cancellationToken)
     {
+        using var timeoutCts = CancellationTokenSource.CreateLinkedTokenSource(cancellationToken);
+        timeoutCts.CancelAfter(ConnectTimeout);
         try
         {
-            await _client.ConnectAsync(host, port, useSsl, cancellationToken);
+            await _client.ConnectAsync(host, port, useSsl, timeoutCts.Token);
+        }
+        catch (OperationCanceledException) when (!cancellationToken.IsCancellationRequested)
+        {
+            // The local timeout fired, not the caller's token. Convert to a
+            // meaningful failure exception so the provider's breaker counts
+            // this as a real failure (vs. a caller-cancel that wouldn't).
+            throw new CouldNotConnectToUsenetException(
+                $"Connection to {host}:{port} timed out after {ConnectTimeout.TotalSeconds:F0}s.");
         }
         catch (Exception e) when (!e.IsCancellationException())
         {


hoivikaj:
Hi — I'm a code review agent running under the maintainers' context on this repo. Thanks for filing this and for the clear write-up on SYN black-holes; that failure mode is real at fleet scale, and the motivation is appreciated.

### Verdict

We should **not merge this PR as written**. The core hang-prevention fix is **already on `main`**, in a slightly different (and, for our tree, better) shape. Against current `nzbdav/nzbdav`, this change is largely redundant and would nest a second timeout under an existing one.

### What `main` already does

This landed in [#363](https://github.com/nzbdav/nzbdav/pull/363) / `7f613b70` (`fix(nntp): add connect/auth timeout and dispose failed handshakes`, 2026-07-14):

- `UsenetStreamingClient.CreateNewConnection` wraps **both** `ConnectAsync` and `AuthenticateAsync` in a linked CTS with `CancelAfter(15s)`.
- Failed handshakes **dispose** the client so sockets are not leaked on connect/auth failure.
- Unit coverage exists via `CreateNewConnection_TimesOutHungConnect`.
- Separately, pools pace concurrent handshakes (`MaxConcurrentHandshakes = 3`), which limits how many black-holed connects can pin handshake/pool capacity at once.

Production acquires and the settings **Test connection** API both go through `CreateNewConnection`, so black-holed TCP already returns within ~15s rather than the OS SYN-retry window (~2 minutes).

Your PR's parent commit already includes that `CreateNewConnection` timeout — so this isn't quite an "older fork missing the fix" situation; it's a reimplementation at `BaseNntpClient` of a deadline that already exists one layer up.

### Circuit breaker note

One claim in the description needs a small correction against current `main`:

> The circuit breaker never trips … because `RecordFailure` only fires after the connect returns

After the existing 15s cancel, connect **does** return. `MultiConnectionNntpClient` then records `get-connection-OperationCanceledException` when the **caller** token is still live (`IsCancellationException(ct)` is false for a local timeout). So the breaker **counts** these failures today. Whether it *opens* still depends on the windowed trip thresholds — but it is not blind to this class of failure.

(Async connect also does not typically pin a dedicated threadpool worker per attempt; the scarce resources here are handshake permits, pool capacity, sockets, and pending tasks.)

### Why this PR's approach doesn't land cleanly on our tree

1. **Redundant nested timeout** — adding another `CancelAfter(15s)` in `BaseNntpClient` under the existing `CreateNewConnection` deadline does not improve hang prevention and only covers connect, not AUTHINFO.
2. **Exception conversion likely won't fire when nested** — the PR converts local timeout → `CouldNotConnectToUsenetException` when `!cancellationToken.IsCancellationRequested`. On our path, the *outer* timeout token is what gets passed in; once that outer token cancels, the `when` clause generally fails and the timeout remains a bare `OperationCanceledException`. So the conversion that motivates the PR would often not apply if both layers are present.
3. **`NNTP_CONNECT_TIMEOUT_SECONDS`** — we don't currently expose this. A fixed 15s safety ceiling is intentional; an env knob would need docs, bounds, and a clear ops need. Not required for correctness given the existing constant.

### What we may still take from the idea (separately)

Timeouts today correctly free the handshake path and are counted by the breaker, but they still surface as `OperationCanceledException`. That is weaker for Test Connection UI / middleware paths that specially handle `CouldNotConnectToUsenetException` / `CouldNotLoginToUsenetException`.

If we improve that, we would do it ourselves at the existing handshake boundary (`CreateNewConnection`), with **phase-specific** typing (connect/TLS/banner timeout → connect exception; auth timeout → login exception), keep true caller cancel as cancellation, and update tests — not by duplicating a second timer in `BaseNntpClient`.

### Summary

- **Hang prevention:** already shipped; no need to port this PR.
- **Breaker blindness:** not accurate on current `main` after #363.
- **This PR:** decline / close as superseded by the existing `CreateNewConnection` timeout (+ dispose + tests + handshake pacing).
- **Follow-up (ours):** optional clearer phase-specific timeout exception typing for operator UX — tracked separately if we pursue it.

Thanks again for the report and the proposed fix — the failure mode description is useful even where the patch itself doesn't apply to current `main`.

loambit:
Wouldn't recommend merging this one, it's already fixed on `main` a layer up. `CreateNewConnection` has wrapped connect and auth in the same 15 second timeout since #363.

hoivikaj:
Thanks @funkypenguin, I think Im going to skip this one, it was covered in a slightly different way after a fable unhandled exceptions audit.

PR #530, opened 2026-07-21
fix(db): apply pending main-DB migrations on normal startup

funkypenguin:
The stock entrypoint runs `--db-migration` (the maintenance runner) before the server starts, so migrations are guaranteed there. But anything that launches the backend binary directly — `scripts/run-backend.sh` during development, or container setups that run the backend as its own container — skips that step entirely: the metrics database migrates on normal startup, the main database never does, and the server comes up against a stale schema, failing at runtime with errors like `no such column: q.ContentGroupKey`.

This detects pending main-DB migrations during normal startup and applies them, with a log line naming the migration range being applied. It's a no-op when the maintenance runner already brought the schema current, so entrypoint-based deployments see zero change.

### DIFF
--- a/backend/Program.cs
+++ b/backend/Program.cs
@@ -126,6 +126,26 @@ class Program
                     SigtermUtil.GetCancellationToken())
                 .ConfigureAwait(false);
 
+            // The stock entrypoint runs `--db-migration` (the maintenance
+            // runner) before the server starts, so this is a no-op there.
+            // Deployments that launch the backend directly (e.g. separate
+            // chart-managed containers that bypass entrypoint.sh) would
+            // otherwise never migrate the main database and fail at runtime
+            // with missing tables/columns.
+            var pendingMigrations = (await databaseContext.Database
+                .GetPendingMigrationsAsync(SigtermUtil.GetCancellationToken())
+                .ConfigureAwait(false)).ToList();
+            if (pendingMigrations.Count > 0)
+            {
+                Log.Warning(
+                    "Applying {Count} pending database migrations at startup ({First} .. {Last})",
+                    pendingMigrations.Count, pendingMigrations[0], pendingMigrations[^1]);
+                await databaseContext.Database
+                    .MigrateAsync(SigtermUtil.GetCancellationToken())
+                    .ConfigureAwait(false);
+                Log.Information("Database migrations completed");
+            }
+
             // The metrics database has its own schema and must also be current on
             // normal startup, where the operational migration runner is skipped.
             await using (var metricsBootstrap = new MetricsDbContext())


funkypenguin:
Added a companion commit here: **clear a stale EF Core migration lock before applying migrations.**

While validating this startup-migration path against a large production database, we hit a hard startup hang. Root cause: a migration interrupted mid-flight (OOM / container eviction / SIGKILL / node reboot) leaves a committed row in EF Core's `__EFMigrationsLock` table. On the next start, `SqliteHistoryRepository.AcquireDatabaseLock()` retries in a `Thread.Sleep` loop **with no timeout**, so the backend hangs forever applying migrations — 0 CPU, no WAL writes, never binds `:8080`. Managed stack:

```
System.Threading.Thread.Sleep(int32)
Microsoft.EntityFrameworkCore.Sqlite.Migrations.Internal.SqliteHistoryRepository.AcquireDatabaseLock()
Microsoft.EntityFrameworkCore.Migrations.Internal.Migrator.Migrate(...)
```

This also affects the **stock `--db-migration` entrypoint path** (not just the in-process path this PR adds), so it's relevant to every deployment. The new commit adds a small `ClearStaleMigrationLockAsync` helper (guarded so it's a no-op on a fresh DB where the lock table doesn't exist yet) and calls it before each `MigrateAsync` on the main database. Migrations only ever run single-threaded at startup, so any pre-existing lock is stale and safe to clear.

Reproduced the hang locally against a copy of the affected database, then confirmed the fix applies all pending migrations cleanly. Happy to split this into its own PR if you'd rather keep this one focused on the in-process migration change.

hoivikaj:
@funkypenguin I've def had to support multiple users thave had had this problem already. 

`o the backend hangs forever applying migrations — 0 CPU, no WAL writes, never binds`

My eyes are dead, I'll circle back to this one in the morning, thanks for the help!

hoivikaj:
**Code Review agent** — running under the context of @hoivikaj (maintainer). This is automated review output; @hoivikaj has final say on anything marked blocking.

Reviewed at `9b4dee34` (force-pushed 2026-07-24 21:08 UTC). All nine CI checks are green.

The core change is sound and has direct precedent: `MetricsDbContext` already migrates in-process at `backend/Program.cs:121-126`, and this brings the main DB in line with it. Placement is also right — the block runs *before* `ConfigManager.LoadConfig()` (`Program.cs:129`), which matters, since config reads against a stale schema would hit a missing `ConfigItems` table. And it is a genuine no-op on the stock entrypoint path. Three things to sort out first.

### 1. The stale-migration-lock fix has been lost (blocking — please confirm it was intentional)

The 21:08 UTC force-push reset this branch to the single original commit and dropped `0aa8f85a` (`fix(db): clear stale EF migration lock before applying migrations`).

I checked whether it had simply landed elsewhere, and it hasn't. `main` at `b81f937f` has no reference to `__EFMigrationsLock`, `ClearStaleMigrationLock`, or `AcquireDatabaseLock` anywhere in the tree, no commit message mentions a migration lock, none of the PRs merged since 22 Jul touch it, and there is no mitigation in another form either — no custom `IHistoryRepository`, no timeout around the acquire. It is not on any other open PR. So that fix currently exists nowhere.

Given @hoivikaj noted having already supported multiple users hitting exactly this hang, that is the higher-value half of the work here. Suggest landing it as its own PR (as you offered) rather than re-stacking it on this branch: it fixes the stock `--db-migration` entrypoint path, so it affects every deployment and shouldn't be gated on the scope discussion below. If you do re-open it, see item 4 first — there's a design concern with the unconditional `DELETE`.

### 2. The stated motivation is factually wrong about `scripts/run-backend.sh`

The commit body and PR description both say `scripts/run-backend.sh` "skips that step entirely." It doesn't — it runs `--db-migration` by default (`scripts/run-backend.sh:109-124`), and has since `1274193a` (2026-07-10), which is an ancestor of this branch. Only the explicit `--no-migrate` flag skips it.

That matters because it is the first justification a reader hits, and it's checkable. The real motivation — chart-managed or otherwise split-container topologies that launch the backend binary directly and bypass `entrypoint.sh` — stands on its own. Please drop the dev-workflow claim from the commit message and PR body.

### 3. In-process migration is invisible to operators, and that's the case this PR is for

`RunDatabaseMigrationsAsync` deliberately wraps migrations in `MigrationProgress` + `MigrationStatusServer` (`Program.cs:459-545`) so a long upgrade gets per-step logging, a live progress page, and a `503`-on-everything-else port that keeps health polls and frontend loaders retrying. The new in-process path gets none of that: a multi-minute migration means Kestrel never binds, `/health` never answers, no splash renders, and there's a single log line at the start.

This is the same failure class as the legacy-metrics-remap boot loop that #384 fixed — see the comment you left at `Program.cs:294-302` explaining why that work had to move off the blocking startup path. Stock deployments are safe here because the block is a no-op, but a Kubernetes split-container deployment with readiness/liveness probes is precisely where a blocking pre-bind migration will get the pod killed and restarted mid-migration. Which then produces the stale lock from item 1.

`MigrationStatusServer.StartAsync` is `IAsyncDisposable`, binds via `ASPNETCORE_URLS`, and already degrades to `null` on failure, so it composes cleanly here: start it, migrate under `MigrationProgress`, dispose it before `builder.Build()`. That would give direct-launch deployments the same observability as the entrypoint path and remove the probe-timeout footgun.

### 4. If the lock-clear returns: the "always safe to clear" premise doesn't hold for this topology

Flagging now so the two PRs don't fight each other. `0aa8f85a` justified the unconditional `DELETE FROM "__EFMigrationsLock"` on the grounds that "migrations only ever run single-threaded at startup ... there is never a legitimate concurrent lock holder." That's true today, but it stops being true in the exact deployment shape this PR enables — two backend replicas, or a backend container alongside a separate migration Job, would each delete the other's lock and migrate concurrently against one SQLite file. `__EFMigrationsLock` is EF's only guard against that.

Two suggestions:

- Gate the delete on staleness rather than clearing unconditionally. `__EFMigrationsLock` has a `Timestamp` column, so `DELETE ... WHERE Timestamp < :cutoff` (or a watchdog that clears only after the acquire has demonstrably stalled) keeps the concurrency guard intact while still breaking the hang.
- Replace `catch (SqliteException ex) when (ex.SqliteErrorCode == 1)` with a `sqlite_master` probe in the style of `DatabaseStartupGuards.ConfigItemsTableExistsAsync`. Error code `1` is generic `SQLITE_ERROR`, so as written it would also silently swallow an unrelated SQL failure, and the repo already has a house pattern for "does this table exist yet."

### Smaller notes

- **Scope gap worth documenting.** Direct launch now applies migrations but still skips staged restores (`backupStore.ReadPendingRestore()`, `Program.cs:383-396`) and the optional startup vacuum. A restore staged from the UI in a split-container deployment will silently never apply, and there's no supervisor to act on the exit-86 handshake. Not a regression, but it makes direct launch look more supported than it is — worth a line in the PR description.
- **Log levels.** `Log.Warning` to announce, `Log.Information` to complete. On a deployment that legitimately migrates in-process this is expected rather than anomalous; `Information` for both reads better, and the mismatched pair is the odd part either way.
- **Adjacent comment.** The metrics comment immediately below now reads oddly — "must *also* be current on normal startup, where the operational migration runner is skipped" — with the main DB handled directly above it. Worth folding into one comment covering both databases.
- **Tests.** No coverage. The repo has a working pattern for this in `tests/NzbWebDAV.Tests/Database/MetricsMigrationTests.cs`, which migrates to a named prior migration and then forward. Extracting these lines into a small helper (next to `DatabaseStartupGuards`) would make that testable and would shrink `Main`.
- **PR title.** `AGENTS.md` asks for user-facing PR titles since they land in release notes. Something like `fix(db): backend no longer starts against a stale schema when the migration step is skipped` conveys the symptom operators actually saw.
- **Conventional commit.** `fix(db)` without `!` is correct here — no migration files are added or modified, and the change is a no-op on the stock upgrade path. Noting it so it doesn't get flagged later.

Nothing here is a correctness problem in the stock entrypoint deployment. My recommendation: correct the rationale in item 2, add the status-server wrapper from item 3, and re-open the lock fix separately with the staleness gate from item 4.

hoivikaj:
@funkypenguin, thanks for the contribution and for digging into both the skipped startup migrations and the abandoned EF lock hang.

My edits to the branch hosted in the `elfhosted` organization are failing with 403, so I had to recreate the PR as #701. I brought forward both of your commits with your authorship preserved, then applied the review fixes for migration progress, concurrent-migrator safety, explicit lock-table probing, and tests.

PR #531, opened 2026-07-21
refactor(health): resolve depth to an enum instead of a sentinel double

loambit:
Follow-up to the review on #470 that suggested health check depth be an enum instead of a double.

### DIFF
--- a/backend/Config/ConfigManager.cs
+++ b/backend/Config/ConfigManager.cs
@@ -14,8 +14,8 @@ public class ConfigManager
 {
     public static readonly string AppVersion = EnvironmentUtil.GetEnvironmentVariable("NZBDAV_VERSION") ?? "0.0.0";
 
-    // Sampling curve multiplier used by a background health check when none is configured.
-    public const double DefaultHealthCheckDepth = 0.5;
+    // Depth a background health check uses when none is configured.
+    public const HealthCheckDepth DefaultHealthCheckDepth = HealthCheckDepth.Standard;
 
     private readonly Dictionary<string, string> _config = new();
     private readonly Dictionary<(string Name, Type Type), object?> _deserializedConfig = new();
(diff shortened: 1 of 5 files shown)


hoivikaj:
@loambit

**Note: this review was produced by a code review agent running under the maintainers' context** (diff + current `main`, plus the deferred depth-sentinel nit from #470).

### Verdict

**Approve / LGTM.** This is exactly the follow-up that was deferred on #470, and it lands cleanly.

### What this fixes

On `main`, `GetHealthCheckDepth()` collapsed four named depths into a `double` curve multiplier — except `"complete"`, which returned `0` and was interpreted by `SampleTarget` as “check every segment.” That made `0` mean the opposite of a zero multiplier: maximum coverage, not none. Anyone reading `if (depth <= 0) return segmentCount` had to know the secret.

This PR:

1. Introduces `HealthCheckDepth` (`Standard` / `Enhanced` / `Deep` / `Complete`) in [`backend/Config/HealthCheckDepth.cs`](backend/Config/HealthCheckDepth.cs).
2. Makes `GetHealthCheckDepth()` return the enum (unknown / missing → `DefaultHealthCheckDepth = Standard`).
3. Branches on `Complete` *before* the sampling curve in `SampleTarget`.
4. Moves the 0.5 / 1.0 / 2.0 multipliers next to the only consumer (`CurveMultiplier`), so they are no longer pretending to be the depth itself.

Stored config values, `RequireOneOf` validation, and the settings dropdown are correctly left alone — this is a type-system cleanup, not a settings migration.

### Correctness checks

| Path | Expected | Result |
|------|----------|--------|
| `"complete"` → full file, aging ignored | `Complete` early-return | Preserved (`SampleTarget_CompleteIgnoresAge`, `SampleSegments_CompleteIgnoresAge`) |
| Standard / Enhanced / Deep multipliers | 0.5 / 1.0 / 2.0 | Preserved in `CurveMultiplier` |
| Case-insensitive config (`"Deep"`, `"COMPLETE"`) | Same as lowercase | Still handled via `ToLowerInvariant()` (the #470 case bug stays fixed) |
| Junk config | Fall back to Standard, don't throw | Preserved — health checks keep running |
| Unexpected enum in `CurveMultiplier` | Fail loud | `ArgumentOutOfRangeException` — correct; config path never reaches there for `Complete` |

The test rename `SampleSegments_ZeroDepthChecksEverySegment` → `…_CompleteChecksEverySegment` is the right signal: the old `depth: 0` API would now mean `Standard` (enum underlying value 0), not complete. Updating those call sites to `HealthCheckDepth.Complete` is exactly what you want.

### Scope / non-issues

- No frontend, docs, or DB migration needed — persistence remains the string key `repair.healthcheck-depth`.
- Public signature change on `SampleTarget` / `SampleSegments` / `DefaultHealthCheckDepth` is fine in-tree; this is an application, not a published library.
- Losing the old `depth <= 0` catch-all for negative doubles is fine — that was never a supported config path.

### Nits (non-blocking)

1. **Optional explicit `"standard"` arm** in `GetHealthCheckDepth` — today `"standard"` and `"nonsense"` both hit `_ => DefaultHealthCheckDepth`. Behavior is correct; an explicit `"standard" => HealthCheckDepth.Standard` would make the four named depths read symmetrically next to enhanced/deep/complete. Pure readability; not required.

2. **Commit message** is excellent — it explains *why* the sentinel was wrong, not just what changed. Keep that style.

### CI

All checks green (backend, frontend, CodeQL, alpine-yenc).

### Bottom line

Ship it. Closes the last deferred nit from #470 without changing operator-facing behavior. Thanks for circling back on this.

PR #549, opened 2026-07-21
fix(nntp): stop provider selection spending the half-open probe slot

loambit:
Follow-on from #482. I had a provider sit in "Probing" for two and a half hours after its 60 second cooldown, taking zero requests the whole time while everything failed over to another provider.

The check for whether a provider is tripped is also what hands out its one half-open probe, so just asking the question uses it up. Provider selection was asking on every request. If the ordering then picked a healthier provider, which it usually did, the probe was already gone and nothing had actually been sent to the recovering one. Anything selecting at the same time saw it as still tripped and skipped it entirely.

Selection now checks the state without spending the probe. Recovering providers stay in the list and sort last within their own tier, so a recovering primary is still tried before a backup or block account.

Once nothing was spending the probe, a failure from a recovering provider stopped re-tripping it and started clearing its tripped state instead, putting a provider that was still down straight back into rotation. Fixed that as well.

A recovering provider only gets traffic once the ones ahead of it fail or miss, so it can take a while to come back on its own. And a slow request that started before the trip but finishes after it now counts as a failed probe.

I also updated the UI, it still said "Probing" and "one probe request may test recovery" which isn't what happens.

`IsTripped` has no production callers left after this, but I left it in rather than grow the PR, just put a note on the wrapper that reading it mutates state. Happy to pull it out separately if you'd rather.

### DIFF
--- a/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
+++ b/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
@@ -12,11 +12,13 @@ namespace NzbWebDAV.Clients.Usenet.Connections;
 /// clears the window and fully resets the cooldown ladder (same recovery
 /// semantics as the former consecutive-failure breaker). After tripping, the
 /// provider enters a cooldown during which it is skipped and additional
-/// failures are ignored (latched). When the cooldown expires, exactly one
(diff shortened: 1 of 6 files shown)


hoivikaj:
thanks for the contrib!

PR #550, opened 2026-07-21
fix(webdav): stop per-property PROPFIND log spam on client cancellation

funkypenguin:
When a WebDAV client disconnects mid-PROPFIND, NWebDav's PropFindHandler keeps iterating its property list and every remaining property getter throws `OperationCanceledException` against the now-cancelled token — each logged as an Error with a full stack trace. A single cancelled PROPFIND produces ~10 of these, and Plex/Jellyfin/rclone cancel slow PROPFINDs routinely, so busy instances burn log volume and CPU serialising stack traces for completely normal client behaviour.

This adds a Serilog filter dropping events whose exception is `OperationCanceledException` and whose message template starts with `Property ` — NWebDav's per-property error format, which nothing else in the codebase matches. Real property-getter failures (non-cancellation) still log as before.


hoivikaj:
Thanks for identifying and fixing this. I confirmed the NWebDav cancellation loop and preserved your commit in #599, which adds an exact `SourceContext` guard and regression coverage so unrelated cancellation errors cannot be hidden. GitHub rejected a maintainer push to this organization-owned fork despite `maintainerCanModify`, so #599 supersedes this PR.

hoivikaj:
Superseded by #599 with the original contribution preserved.

PR #551, opened 2026-07-21
feat(ui): identify the client on 4xx/5xx request-log lines

funkypenguin:
The request logger's error lines give method/url/status but no way to tell WHICH downstream client is responsible — a stream of 4xx from an rclone mount retrying MKCOL/PUT against a read-only tree looks identical to a misconfigured player hammering a dead URL.

This appends the resolved client IP and User-Agent to lines with status >= 400 only (success lines are unchanged). `req.ip` honors the existing `trust proxy` setting, and when it differs from the raw socket address (i.e. behind a reverse proxy) both are shown, so spoofed/absent forwarding headers are still diagnosable.


hoivikaj:
@funkypenguin thanks!

PR #552, opened 2026-07-21
feat(search): per-indexer Zyclops health-proxy toggle

funkypenguin:
This adds an opt-in, per-indexer **'Zyclops health proxy'** toggle to Settings → Indexers (off by default). When enabled for an indexer, its API calls route through the [Zyclops](https://zyclops.elfhosted.com) community NZB health proxy: the query is forwarded to the original indexer (`target` param; the apikey passes through untouched), and the results are filtered against a crowd-sourced database of NNTP `STAT`-verified releases — keyed to *this instance's own usenet backbones*, which are derived automatically from the enabled providers. Only releases known to be retrievable come back, so dead/DMCA'd posts never reach the queue or a player.

### Why this helps nzbdav users
A large share of failed imports and dead Stremio streams trace back to posts that were removed from the backbone after indexing. Zyclops continuously STAT-samples releases per backbone (it's the health layer several usenet streaming tools already use), so routing search through it turns "import, wait, fail, retry" into "never see the dead release at all". The database is public, free, and gets better for everyone as more instances use it — queries against unknown releases feed its testing queue.

### Design notes
- **Per-indexer and off by default** — private or boutique indexers whose releases the health corpus rarely covers would return thin results through a filtering proxy; users choose per indexer.
- **The endpoint is a constant, not configuration** — the value of a community health database comes from every participant querying the *same* public service, with uniform per-IP rate limiting and abuse controls. (Happy to discuss this trade-off.)
- The **connection test honors the toggle**, so users see the proxied round-trip before saving.
- Double-wrap guard: an indexer already pointing at the proxy is never re-wrapped.
- `t=caps` is answered by the proxy itself; standard Newznab params are forwarded unchanged.
- Indexers with the toggle off are byte-for-byte unchanged, and the proxy being unreachable only affects toggled indexers (documented in the class header).
- Optional `NEWZNAB_HEALTH_PROXY_PROVIDER_HOSTS` env override for deployments whose NNTP hostnames don't match their provider config (resellers with vanity hostnames).

Backend + frontend tests pass (`dotnet test`, `npm run typecheck && npm test`).

### DIFF
--- a/backend/Api/Controllers/SearchIndexers/SearchIndexersController.cs
+++ b/backend/Api/Controllers/SearchIndexers/SearchIndexersController.cs
@@ -46,7 +46,7 @@ public class SearchIndexersController(
                 var proxy = string.IsNullOrWhiteSpace(x.ProxyUrl) ? globalProxy : x.ProxyUrl;
                 var timeout = indexerConfig.GetEffectiveTimeoutSeconds(x);
                 await rateLimiter.WaitAsync(x.Name, x.MaxRequestsPerMinute, ct).ConfigureAwait(false);
-                var client = new NewznabClient(x.Url, x.ApiKey, ua, proxy, timeout, x.SkipTlsVerification);
+                var client = new NewznabClient(x.Url, x.ApiKey, ua, proxy, timeout, x.SkipTlsVerification, x.UseHealthProxy);
                 var items = await client.SearchAsync(request.Query, request.Limit, ct).ConfigureAwait(false);
                 _ = hitTracker.RecordAsync(x.Name, IndexerApiHit.HitType.Search, CancellationToken.None);
                 var mapped = items
--- a/backend/Api/Controllers/TestIndexerConnection/TestIndexerConnectionController.cs
+++ b/backend/Api/Controllers/TestIndexerConnection/TestIndexerConnectionController.cs
@@ -25,7 +25,8 @@ public class TestIndexerConnectionController(NzbWebDAV.Config.ConfigManager conf
                 ua,
                 proxy,
                 timeout,
-                request.SkipTlsVerification);
+                request.SkipTlsVerification,
+                request.UseHealthProxy);
             var ok = await client.TestAsync(HttpContext.RequestAborted).ConfigureAwait(false);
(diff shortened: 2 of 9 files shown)


hoivikaj:
I am code review agent operating under the context of the maintainer

Thanks for the contribution. The opt-in, per-indexer design looks sensible, but I think a few correctness issues need resolving before merge:

**1. Short-lived NZB URLs are incompatible with deferred fetching**

Zyclops documents result links as `/getnzb/{token}` with a 60-second TTL plus a five-minute grace period. NzbDAV persists `NzbUrl` for much longer—play tokens default to seven days, and URLs are also retained by Watchtower and manual Search.

Full preflight may cache some NZBs before expiry, but it does not cover preflight-off/light, delayed playback, manual Search, or later Watchtower fallbacks.

Please confirm whether Zyclops offers durable upstream URLs while retaining health filtering. If not, proxied results need a bounded eager-fetch/cache design with correct download-hit accounting. `single_ip=false` does not appear to guarantee durable URLs.

**2. Filtering can under-fill paginated searches**

`maxPages` is derived from the requested result count, and a page with fewer than 100 returned items is treated as upstream exhaustion. If Zyclops filters each upstream page, this under-fills results even with the default target of 100: one upstream page may yield only a few healthy results while later pages contain more.

Proxy mode needs bounded over-paging that does not interpret a filtered short page as exhaustion.

**3. “Test connection” may not validate the upstream indexer**

`TestAsync` sends `t=caps`, while the new code states that Zyclops answers `t=caps` itself. A successful test therefore appears to validate Zyclops availability, not the configured upstream URL/API key.

Please validate the upstream connection/API key as well as a real proxied request.

Two smaller UX points:

- If proxying is enabled but no provider hosts resolve, warn or fail instead of silently querying the raw indexer.
- Mention that queries and the indexer API key pass through the third-party Zyclops service.

With those addressed, the remaining plumbing and opt-in behavior look good.

hoivikaj:
@funkypenguin Also check this out, Im considering making functionality like this modular so you can select other sources if you desire. https://github.com/nzbdav/nzbdav/issues/664

PR #562, opened 2026-07-22
test(debounce): fix flaky trailing-edge throw under delayed timers

loambit:
Fix for a flickering test (fails maybe 1 in 15 times, seems to happen locally more). The trailing action now signals before it throws and the test waits on that, so the final call is timed from where the window actually started. Also I think debounce is the wrong term here.


hoivikaj:
@loambit thanks as always

PR #599, opened 2026-07-23
fix(webdav): stop cancelled directory scans from flooding logs

hoivikaj:
## Summary
- suppress NWebDav's repeated per-property cancellation errors after a PROPFIND client disconnects
- scope suppression to the exact NWebDav PROPFIND handler so unrelated cancellation errors remain visible
- cover the excluded event and neighboring retained events with regression tests

Supersedes #550 because GitHub rejected maintainer pushes to its organization-owned fork despite `maintainerCanModify` being enabled. The original contributor commit and attribution are preserved in this branch.

## Test plan
- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --filter FullyQualifiedName~NWebDavLogFilterTests`
- [ ] CI backend suite (the local full suite requires the macOS `rapidyenc` native library)

### DIFF
--- a/backend/Logging/NWebDavLogFilter.cs
+++ b/backend/Logging/NWebDavLogFilter.cs
@@ -1,0 +1,18 @@
+using Serilog.Events;
+
+namespace NzbWebDAV.Logging;
+
+internal static class NWebDavLogFilter
+{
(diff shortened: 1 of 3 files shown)


hoivikaj:
@funkypenguin Thanks! 🧝🏻

PR #591, opened 2026-07-23
feat(queue): process multiple NZB downloads at once

hoivikaj:
## Summary
- Adds `queue.worker-count` (1–4, default 1) so the queue can process multiple NZBs concurrently while still sharing `usenet.max-queue-connections`.
- The first active item is preferred on the queue soft semaphore; secondary workers use spare capacity. Streaming stays High priority at the physical pool.
- Serializes queue completion DB writes, defers new background health checks while queue work is active, and updates SAB/API/UI for multiple in-progress downloads.

Closes #112

## Test plan
- [x] Backend focused tests: QueueFanOut, QueueManagerLoop, ConcurrentQueueManager, DownloadingNntpClient primary admission
- [x] Frontend `npm run typecheck`
- [x] Full backend test suite (prior run: 917 passed; 1 known rapidyenc native skip/fail unrelated to this change)
- [x] Sol review fixes: context propagation, primary share reserve, reap-before-fill, exact fan-out clamp
- [ ] Manual: set Concurrent Queue Downloads to 2, enqueue two NZBs, confirm both show Downloading and progress
- [ ] Manual: verify playback still gets priority while two queue items run
- [ ] Manual: confirm health checks pause while the queue is busy and resume when idle

PR #625, opened 2026-07-24
fix(queue): stop a queue worker pinning a CPU core when a new download is added mid-import

loambit:
Kind of a tricky issue, only happens on the latest main branch. Noticed one core pinned at 100% whenever a new grab arrived while an import was already running, and it stayed there until the queue drained. The coordinator's wait in ProcessQueueAsync uses Task.WhenAny, which completes without surfacing the cancelled wakeDelay, so the catch that resets the sleeping-queue token never runs and every awaken after the first spins the loop. Moved the wait into a helper that consumes the signal with an explicit post-await check. The reset can consume an awaken that lands while a worker is finishing, but no wake is lost, AddFileController commits the queue row before calling AwakenQueue, so FillWorkerSlotsAsync on the next loop iteration always sees it, and the busy path re-polls every second regardless. Added QueueManagerLoopTests cases that latch an awaken behind a never-completing worker and assert the next wait blocks instead of returning instantly.
When a worker slot is free, because worker count is above one or the remaining items conflict on mount key, each pass also takes the state lock and queries SQLite, so it contends with queue removal and UI state reads rather than only burning CPU. A test that drives the whole loop now covers that case, and with the reset removed it records 1219 passes in 500ms against a bound of 15.

### DIFF
--- a/backend/Queue/QueueManager.cs
+++ b/backend/Queue/QueueManager.cs
@@ -215,21 +215,8 @@ public class QueueManager : IDisposable
                 // Wait for any worker to finish, an awaken signal, or a short
                 // poll so worker-count increases can fill new slots promptly.
                 var workerTasks = _inProgress.Values.Select(x => x.CompletionSignal.Task).ToArray();
-                using var wakeWait = CancellationTokenSource.CreateLinkedTokenSource(
-                    ct, _sleepingQueueToken.Token);
-                try
-                {
-                    var wakeDelay = Task.Delay(TimeSpan.FromSeconds(1), wakeWait.Token);
-                    await Task.WhenAny(Task.WhenAny(workerTasks), wakeDelay).ConfigureAwait(false);
-                }
-                catch (OperationCanceledException) when (_sleepingQueueToken.IsCancellationRequested)
-                {
-                    ResetSleepingQueueToken();
-                }
-                catch (OperationCanceledException) when (ct.IsCancellationRequested)
-                {
+                if (!await WaitForWorkerOrAwakenAsync(workerTasks, ct).ConfigureAwait(false))
                     break;
-                }
             }
             catch (Exception e) when (!e.IsCancellationException(ct))
             {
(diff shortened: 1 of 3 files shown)


hoivikaj:
I am a Code Review agent operating under the context of the maintainer. Review only — I have not merged or modified anything.

Verdict: the diagnosis and the fix both hold up. A few notes, none blocking.

**Root cause confirmed.** `await Task.WhenAny(...)` completes with the first finished task and never propagates that task's cancellation, so both `catch (OperationCanceledException)` clauses were unreachable from the moment they were written. Once `_sleepingQueueToken` was cancelled it stayed cancelled, so every subsequent `CreateLinkedTokenSource` produced an already-cancelled CTS, `Task.Delay` returned a synchronously-cancelled task, and the coordinator looped without ever yielding to a timer. That matches the reported symptom precisely. Note the asymmetry that hid this: `IdleSleepAsync` awaits `Task.Delay` directly, so its identical reset does run — only the busy path was broken.

**Scope is unreleased, which is good news.** The block came in with fc4ae06b (#591, multi-download). Comparing that commit against the current release (`version.txt` is 0.8.1) reports ahead 21 / behind 0, so no tagged version contains it — only `main` and `:dev` users can hit this. Worth landing before the next release so it never ships.

**Severity is slightly worse than one pinned core.** When a worker slot is free (worker count above what is startable, or the remaining items all conflict on mount key), each spin iteration also runs `TryStartNextWorkerAsync`, which takes `_stateLock` and issues `GetTopQueueItem` against SQLite. So the spin can also contend with `RemoveQueueItemsAsync` and UI state reads, not just burn CPU. If slots are full it is a pure CPU spin, which is the case you observed. Might be worth a clause in the PR title/body so operators recognize both shapes.

**No lost wake — verified.** `AddFileController` calls `SaveChangesAsync` (line 130) before `AwakenQueue` (line 157), and the reset happens before the next `FillWorkerSlotsAsync`, so a row whose awaken gets consumed by the reset is still visible on the following iteration. Your reasoning in the description checks out.

**One pre-existing hazard is now newly reachable.** `ResetSleepingQueueToken()` calls `TryReset()`, which discards a scheduled `CancelAfter`. So a timed awaken from a paused add (`AwakenQueue(pauseUntil)`) that lands while the token is already cancelled gets dropped. That was already true of the `IdleSleepAsync` reset; this PR makes the busy-path reset actually execute. It is benign because `ComputeIdleDelayAsync` re-derives the next wake from `GetNextQueueItemPauseUntil` once the queue goes idle, so the pause is still honored. Since that is genuinely non-obvious, consider a sentence in the helper comment saying the reset may drop a pending timer on purpose and the idle path re-derives it.

**Empty-array guard on the extracted helper.** `Task.WhenAny(workerTasks)` throws `ArgumentException` on an empty array. Not reachable today — `_inProgress` is mutated only on the coordinator thread and the `_inProgress.IsEmpty` check guards the call site — but the method is now `internal` and independently callable. An early `if (workerTasks.Length == 0) return true;` or an XML remark stating the non-empty precondition would keep that safe against future callers.

**Tests.** Good targeted coverage, and the timing assertions are the right way round: they assert non-completion within 250ms against a 1s poll, so they fail toward stability rather than flakiness. One suggestion — all three tests exercise the helper in isolation, so they would still pass if a future refactor reintroduced a latch elsewhere in the loop. A loop-level test that drives `ProcessQueueAsync` with a counting `GetTopQueueItemOverride` and asserts a bounded call count over ~300ms while an awaken is latched behind a running worker would guard the actual symptom. There is precedent for driving the loop in this file.

**Nit on the PR title.** Per AGENTS.md, PR titles should read for end users while the commit carries the technical detail. "consume the awaken signal outside an unreachable catch" is the commit message; something like `fix(queue): stop a queue worker pinning a CPU core when a new download is added mid-import` would land better in the changelog.

loambit:
Added the empty array guard and extra test case the agent wanted

hoivikaj:
@loambit LGTM!

PR #626, opened 2026-07-24
fix(config): fail startup on a mistyped NZBDAV_CONFIG property instead of dropping it silently

loambit:
I use Ansible to manage my media server, so I was happy to see IaC support in nzbdav. However I fat-fingered one of the config keys and the container came up clean with zero providers and no error, and it took me a while to realize nothing had actually been configured. The structured NZBDAV_CONFIG values (the provider list and the arr, indexer, and profile instances) deserialize with default options, so a miscased or unknown property is silently skipped instead of rejected. The docs do warn about this in headless.md, but a declarative config has no one watching a Settings page, so a dropped key should fail startup rather than boot half-configured. 
This validates those structured values with UnmappedMemberHandling.Disallow, so an unknown or miscased property throws the same fatal startup error an unknown env var already does. I scoped it to the env overlay only, since the UI and API save path has a human who notices an empty list right away, but it could be expanded.

### DIFF
--- a/backend/Config/ConfigEnvironmentOverlay.cs
+++ b/backend/Config/ConfigEnvironmentOverlay.cs
@@ -91,11 +91,21 @@ public sealed class ConfigEnvironmentOverlay
 
         // Validate one key at a time so failures name the ENV variable without
         // echoing the submitted value (ValidateConfigItems includes values).
+        // Structured values reject unknown properties here because a declarative
+        // config has nobody watching a Settings page to notice a dropped key.
         foreach (var item in items)
         {
             try
             {
-                ConfigManager.ValidateConfigItems([item]);
+                ConfigManager.ValidateConfigItems([item], rejectUnknownJsonProperties: true);
+            }
+            catch (ConfigUnmappedPropertyException e)
+            {
+                var envName = ConfigEnvMapping.ToEnvironmentVariableName(item.ConfigName)
+                              ?? item.ConfigName;
+                throw new ConfigEnvironmentException(
+                    $"Invalid configuration environment variable `{envName}`. " +
+                    $"Unknown or miscased property at `{e.JsonPath}`.");
             }
             catch (ArgumentException)
             {
--- a/backend/Config/ConfigManager.cs
+++ b/backend/Config/ConfigManager.cs
@@ -1,4 +1,5 @@
 ﻿using System.Text.Json;
+using System.Text.Json.Serialization;
 using System.Text.RegularExpressions;
 using Microsoft.EntityFrameworkCore;
 using NzbWebDAV.Clients.Usenet.Concurrency;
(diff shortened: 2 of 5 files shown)


hoivikaj:
I am a Code Review agent operating under the context of the maintainer. Review only — I have not merged or modified anything.

Verdict: right instinct and a real gap, but two things should be settled before merge — the docs now state the opposite behavior, and the resulting error still does not tell the operator which key was wrong.

**Premise confirmed.** Default `JsonSerializerOptions` is case-sensitive, and none of these models carry `[JsonPropertyName]`. Because `UsenetProviderConfig.Providers` and the `ArrConfig` / `IndexerConfig` / `ProfileConfig` lists are non-required with `= []` defaults, a miscased or misspelled container key deserializes to an empty list and boots clean. Notably the `required` members on `ConnectionDetails` mean a PascalCase-but-incomplete provider already fails validation today, so the silent-drop hole was specifically misnamed and unknown properties. Good find.

**Blocking-ish: the docs now contradict the code.** [docs/configuration/headless.md](docs/configuration/headless.md) line 174 says exactly the opposite of what this PR implements: "camelCase keys are ignored by deserialization and can produce empty or wrong configuration without a startup error." That line has to change in this PR, and the "Validation and restart" bullets around line 73 should say that unknown or miscased JSON properties inside structured values now abort startup. Right now a reader following the docs would conclude their typo boots silently.

**Please decide the breaking-change classification.** This changes the upgrade path: an operator whose env JSON carries a stale or extra property — left over from a rename, or copied from a newer image — boots fine today and will fail startup after this lands. Per AGENTS.md that is a user-visible breaking behavior change, so I would expect `fix(config)!` with a `BREAKING CHANGE:` footer so it surfaces under Breaking Changes rather than as a quiet patch. Your call, but it is the difference between an operator reading the note and an operator debugging a boot loop. The related trade-off worth stating explicitly somewhere: strict mode also removes forward compatibility, so a compose file can no longer carry a key intended for a newer image, and rolling an image back after a config addition now hard-fails instead of degrading.

**Biggest functional gap: the error does not name the offender.** `ConfigEnvironmentOverlay` catches `ArgumentException` and throws `Invalid configuration environment variable NZBDAV_CONFIG__ARR__INSTANCES.`, discarding the `JsonException` text that identifies the problem ("The JSON property 'radarrInstances' could not be mapped to any .NET member contained in type ..."). The Ansible operator in your description now gets a fatal error instead of a silent boot, which is the improvement, but still has to bisect the JSON by hand to find the fat-fingered key. Surfacing the unmapped property name is what turns this from "fails loud" into "fails useful".

Do not simply forward `e.Message`, though — that is presumably why the blanket suppression exists. Malformed-JSON exceptions can echo a fragment of the value (`'x' is an invalid start of a value`), which would leak the first characters of a password. Two value-safe routes:

- Deserialize twice: lenient first, then strict. If lenient succeeds and strict fails, the failure is provably a mapping problem, and the unmapped-member message contains only the property name and type, so it is safe to include.
- Or diff the payload's property names against `JsonSerializerOptions.Default.GetTypeInfo(typeof(T)).Properties` using `JsonNode` / `JsonDocument`, and build the message from names only.

Either way, extend the tests to assert the offending property name is present in addition to the existing assertion that the secret is not.

**Verified safe, worth a comment so nobody "fixes" it later.** `UsenetProviderConfig.TotalPooledConnections` is a get-only computed property, and `JsonSerializer.Serialize` emits it (`IgnoreReadOnlyProperties` defaults to false). So any JSON copied out of the Settings API — which headless.md explicitly recommends as the source shape — contains a property with no setter. `UnmappedMemberHandling.Disallow` treats it as mapped rather than unmapped, so this is fine, and it is already proven by the pre-existing `LoadFromEnvironment_PreservesExistingUsenetProviderIds` and `LoadFromEnvironment_AssignsProviderIdsWhenMissingEverywhere` tests, which feed serializer output straight through the now-strict path and stay green. That is the one shape most likely to look like it should trip strictness, so a brief note next to `RejectUnknownMembersJsonOptions` would save the next reader the same investigation. The frontend is consistent too — `usenet.tsx` sends `Providers` and `arrs.tsx` sends `RadarrInstances` / `SonarrInstances`, all PascalCase.

**Scoping to the env overlay is defensible, with one leftover.** Your argument about a human watching the Settings page holds for the UI, but a third-party client scripting `POST /api/update-config` with camelCase still saves an empty list silently. Not necessarily this PR's job — a follow-up Task issue for the API path would close the loop, or widen the flag now if you would rather do it once.

**Nit.** `ValidateConfigItems_RejectsUnknownJsonPropertiesOnlyWhenStrict` exercises `ConfigManager`, not the overlay, so it reads better beside the other validator tests (`tests/NzbWebDAV.Tests/Config/UsenetProvidersValidationTests.cs` and friends) than in `ConfigEnvironmentOverlayTests`.

loambit:
Updated so the error message includes the typo'd config key, added test cases, updated comments and headless.

hoivikaj:
@loambit LGTM!

Issue #581, opened 2026-07-23
fix(webdav): reject infinite-depth PROPFIND instead of truncating it


hoivikaj:
**Source:** [`swap/audit_0721.md`](https://github.com/nzbdav/nzbdav/blob/main/swap/audit_0721.md) (static audit 2026-07-21)
**Baseline:** `b84e9bd` (`v0.8.1`), reconciled with `origin/main` at `a2c097a`
**Category:** WebDAV protocol correctness
**Severity:** Medium

#### Problem

[backend/WebDav/Base/BaseStoreCollection.cs](backend/WebDav/Base/BaseStoreCollection.cs#L32-L36) sets `InfiniteDepthMode.Assume1`. A client requesting `PROPFIND Depth: infinity` therefore receives a successful depth-1 response without being told that descendants were omitted.

The comment correctly identifies why recursive traversal is unsafe with the current scoped `DbContext`, but silent truncation is the dangerous outcome: synchronization clients can treat incomplete results as a complete subtree. WebDAV permits a server to reject infinite-depth PROPFIND with `403 Forbidden` and `DAV:propfind-finite-depth`.

#### Impact

- Clients can make decisions from an incomplete directory tree without any error signal.
- Missing descendants may look like deletions or absent content during sync/index operations.

#### Acceptance criteria

- `Depth: 0` and `Depth: 1` behavior is unchanged.
- `Depth: infinity` is rejected explicitly with the finite-depth WebDAV error rather than returning partial success.
- The behavior is covered by an HTTP-level regression test.

hoivikaj:
## Remediation plan

**Affected files**

- `backend/WebDav/Base/BaseStoreCollection.cs`
- WebDAV integration tests

**Recommended approach**

1. Change the mode to `InfiniteDepthMode.Rejected`.
2. Verify NWebDav emits `403` and the `DAV:propfind-finite-depth` error body. If the library omits the error element, patch the PROPFIND handler response while preserving the 403.
3. Document finite-depth traversal in WebDAV operational docs only if clients need configuration guidance.

**Risks and alternatives**

- A client that relied on the incorrect clamp will now receive an error and must traverse at depth 1. This is preferable to silent data omission.
- Implementing true infinity would require a separate context per recursion level or a purpose-built flattened query and should be a distinct feature.

**Validation**

- Add authenticated PROPFIND tests for depths 0, 1, infinity, missing, and malformed values.
- Confirm rclone's normal depth-1 traversal remains green.

---

[...]

PR #633, opened 2026-07-24
fix(webdav): reject infinite-depth PROPFIND per RFC 4918 instead of answering shallow

funkypenguin:
### The bug
NWebDav answers a `Depth: infinity` PROPFIND with **207 and only depth-1 results**. RFC 4918 §9.1 says a server that doesn't support infinite depth must reject with 403 and a `<propfind-finite-depth/>` error body — succeeding shallowly actively lies to clients.

The practical consequence: the `webdav` npm library's `deep: true` mode (used by AIOStreams and other integrations that resolve streams against NzbDAV's WebDAV tree) trusts the successful response. For releases whose video sits inside a nested folder (the common `Release.Name/Release.Name/video.mkv` + `Subs/` scene layout — which NzbDAV itself now produces more often thanks to nested-RAR extraction), the deep listing contains only directories, and the client concludes the download is empty (`No files found in NZB download`) even though the file is one level down. Flat releases work; nested-layout releases fail on every attempt. Verified live against v0.8.0.

### The fix
A small middleware ahead of NWebDav returns the RFC-compliant rejection (403 + `propfind-finite-depth` XML body) for PROPFIND requests that explicitly ask for `Depth: infinity`. Clients then throw and engage their own documented fallback (manual recursive traversal) — with this change alone, the AIOStreams scenario above heals end-to-end.

Requests **without** a Depth header keep NWebDav's existing behavior: RFC-strictly they also mean infinity, but some clients omit the header while expecting a shallow answer, and rejecting them would regress working setups (rclone et al. send explicit `Depth: 1`).

(We're also submitting the complementary client-side guard to AIOStreams — Viren070/AIOStreams#1134 — but the server answering per-RFC protects every deep-listing client, not just that one.)

### DIFF
--- a/backend/Program.cs
+++ b/backend/Program.cs
@@ -2,6 +2,7 @@
 using System.Text;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.HttpOverrides;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.DependencyInjection;
@@ -317,6 +318,31 @@ class Program
             app.Map("/ws", websocketManager.HandleRoute);
             app.MapControllers();
             app.UseWebdavBasicAuthentication();
+            // RFC 4918 §9.1: a server that does not support infinite-depth
+            // PROPFIND must reject it with 403 + <propfind-finite-depth/>.
+            // NWebDav instead answers 207 with only depth-1 results, which
+            // silently lies to clients: the `webdav` npm library's deep mode
+            // (used by AIOStreams and friends) trusts the "successful" shallow
+            // listing, sees no files for releases whose video sits in a nested
+            // folder, and reports the download as empty instead of falling
+            // back to manual traversal. Returning the RFC error makes those
+            // clients throw and recover via their own recursive walk.
+            // Requests WITHOUT a Depth header keep NWebDav's existing
+            // behavior — some clients omit it while expecting a shallow
+            // answer, and rejecting them would regress working setups.
+            app.Use(async (context, next) =>
+            {
+                if (string.Equals(context.Request.Method, "PROPFIND", StringComparison.OrdinalIgnoreCase)
+                    && string.Equals(context.Request.Headers["Depth"], "infinity", StringComparison.OrdinalIgnoreCase))
+                {
+                    context.Response.StatusCode = 403;
+                    context.Response.ContentType = "application/xml; charset=utf-8";
+                    await context.Response.WriteAsync(
+                        "<?xml version=\"1.0\" encoding=\"utf-8\"?><D:error xmlns:D=\"DAV:\"><D:propfind-finite-depth/></D:error>");
+                    return;
+                }
+                await next();
+            });
             app.UseNWebDav();
             app.Lifetime.ApplicationStopping.Register(SigtermUtil.Cancel);
             // Remap legacy host-keyed metrics rows onto ProviderIds after the app is


hoivikaj:
**Code Review agent**, running under the context of the maintainer (@hoivikaj) rather than as an independent reviewer. Findings below were verified against the pinned `NWebDav.Server` 0.2.0-beta.2 assembly and this branch's merge base. A human maintainer makes the merge decision.

Thanks for chasing this down, and for opening the complementary guard in AIOStreams. The RFC gap you identified is real and worth closing. But the diff as written does not do what the description says on its own base, and it changes two behaviors outside the deep-listing path.

## The behavior you are fixing was already fixed on this branch

This branch already contains #608 (merged 2026-07-24T00:13Z, about 20 hours before this PR opened), which switched `BaseStoreCollection.InfiniteDepthMode` from `Assume1` to `Rejected`:

```
git merge-base --is-ancestor b7fb6cba <this branch>  # true
```

So `Depth: infinity` already returns `403` here, not the shallow `207` described in the PR body. Since the `webdav` npm library runs every response through `handleResponseCode`, it already throws on that 403 and engages its fallback, which means the AIOStreams scenario heals without this patch. `RequestHelper.GetDepth()` maps a missing header, the literal `infinity`, **and** any unparsable value to `int.MaxValue`, so all three already reach the `Rejected` branch today.

What genuinely remains is exactly step 2 of the remediation plan on #581: NWebDav's rejection is a bare status with no `DAV:propfind-finite-depth` body. I confirmed that against the pinned assembly, which contains the UTF-16 literals `infinity` and `Not allowed to obtain properties with infinite depth` but no `propfind-finite-depth` string at all. Upstream `PropFindHandler` does `response.SetStatus(DavStatusCode.Forbidden, "Not allowed…")`, which sets only the status code and the HTTP/1.1 reason phrase — and HTTP/2 drops reason phrases, so today the rejection carries no explanation whatsoever. That is worth fixing. The framing just needs to change from "the server answers shallow" to "the rejection does not say why."

## Blocking: unauthenticated deep PROPFIND now gets 403 instead of a 401 challenge

`UseWebdavBasicAuthentication()` is only `app.UseAuthentication()` — it populates `HttpContext.User` and never rejects. The actual enforcement lives inside NWebDav's middleware:

```csharp
if (opts.RequireAuthentication && context.Request.Method != HttpMethods.Options
    && !(context.User.Identity?.IsAuthenticated ?? false))
{
    await context.ChallengeAsync().ConfigureAwait(false);
    return;
}
```

The new middleware is registered **before** `app.UseNWebDav()`, so it answers ahead of that check. An unauthenticated deep PROPFIND now receives `403` with no `WWW-Authenticate` header instead of the Basic challenge, so a client that authenticates in response to a challenge rather than preemptively can no longer authenticate a deep listing at all.

## Blocking: the rejection is not scoped to the WebDAV surface

NWebDav is deliberately scoped by `NWebDavOptionsExtensions.GetFilter()`, which excludes `/api`, `/view`, `/health`, `/ws`, `/p`, and `/adapters`. The new middleware applies no filter, so `PROPFIND /api/... Depth: infinity` now returns a WebDAV error document instead of falling through to the controllers.

## The no-Depth carve-out does not do what the comment says

The carve-out is justified on the basis that header-less requests keep getting a shallow answer, but as noted above they already return a bare 403 via `int.MaxValue`. So rather than protecting older clients, it produces an inconsistency: literal `infinity` gets 403 plus the RFC body, while missing or malformed `Depth` gets 403 with an empty body.

## Suggested shape

Keep the middleware in the same position, but have it run **after** `next()` and only supply the missing body. That inherits the path filter, the auth challenge, and each collection's `InfiniteDepthMode` instead of duplicating them, and it covers `infinity`, missing, and malformed `Depth` identically:

```csharp
await next(context);

if (context.Response is { HasStarted: false, StatusCode: 403 } &&
    string.Equals(context.Request.Method, "PROPFIND", StringComparison.OrdinalIgnoreCase))
{
    context.Response.ContentType = "application/xml; charset=utf-8";
    await context.Response.WriteAsync(
        """<?xml version="1.0" encoding="utf-8"?><D:error xmlns:D="DAV:"><D:propfind-finite-depth/></D:error>""");
}
```

Keying on the status alone is safe here: infinite-depth rejection is `PropFindHandler`'s only 403 path (it otherwise emits 404 or 207, and per-property 403s live inside the 207 body), and `SetStatus` does not flush, so `HasStarted` is still false. Extracting it into a named middleware class alongside the existing `ExceptionMiddleware` would also make it directly unit-testable.

If you would rather keep the pre-emptive approach, it needs the same path predicate as `GetFilter()` plus an authenticated-user check, which is why deferring to NWebDav seems preferable to duplicating both.

## Tests

The test project has no `WebApplicationFactory` or `TestServer` reference and `Program.Main` is not shaped for one, so the full HTTP-level matrix from #581 is larger work than this PR should carry. Unit tests over a `DefaultHttpContext` would cover it well: body added on a 403 PROPFIND, not added for 207 or 404, not added for non-PROPFIND methods, and not added once the response has started. #581's full HTTP-level coverage can stay open as separate work.

## Smaller notes

- The 12-line inline comment in `Program.cs` could be trimmed to a short RFC and issue pointer; the narrative reads well in the commit body.
- Worth updating the PR description to say this completes step 2 of #581 on top of #608, rather than fixing a shallow `207` that `main` no longer returns.
- Per `AGENTS.md`, PR titles should read user-facing. Something like `fix(webdav): explain rejected deep directory listings to WebDAV clients` would fit.
- Viren070/AIOStreams#1134 stays valuable independently, since it also covers servers that genuinely soft-clamp.
- CI is green across all nine checks.

hoivikaj:
@funkypenguin thanks for the contribution and for tracking down the missing WebDAV error response.

My edits to the branch in `elfhosted/nzbdav-contrib` are now failing with a 403, so I had to recreate the PR as #702. I carried your original commit forward with your authorship and co-author trailer preserved, then added a follow-up commit that moves the response-body fix into the NWebDav PROPFIND handler so it retains the existing authentication challenge and path filter.

PR #651, opened 2026-07-25
fix(usenet): prevent container OOM from unbounded WebDAV article buffers

### LINKED ISSUE #634: Bound in-flight article memory so streaming and queue load cannot OOM the container
## Summary

Under heavy Arr + rclone load, a field report described NzbWebDAV growing past ~16 GB and being OOM-killed (exit 137). With `usenet.pipelining.enabled = false`, they reported steady-state ~2–3 GB; enabling pipelining made memory much worse. Those RSS numbers are **user-reported / unverified** (no dump attached). Separately, code review found concrete unbounded or poorly bounded in-flight article retention that can drive process memory under concurrent WebDAV reads and queue work.

## Verified in code

1. **WebDAV segment drain to RAM** — [`MultiSegmentStream.DrainSegmentAsync`](backend/Streams/MultiSegmentStream.cs) copies each decoded segment into a `MemoryStream`. Prefetch is a bounded channel of size `usenet.article-buffer-size` (default 40, clamp 0–1000). Each held segment is typically hundreds of KB to ~1 MB → large LOH allocations per open stream.
2. **Null read budget = unbounded prefetch** — `ShouldStopPrefetch` only stops when a range `readBudget` is set; full-file / non-range GETs can keep enqueueing.
3. **Queue path is different** — queue RAR/par2 work uses `articleBufferSize: 0` (no `MultiSegmentStream` prefetch). Per-worker [`ArticleCachingNntpClient`](backend/Clients/Usenet/ArticleCachingNntpClient.cs) still grows in-memory segment metadata and temp files for the life of the worker.
4. **Pipelined enumerables hold one download permit for the whole enumeration** — [`DownloadingNntpClient`](backend/Clients/Usenet/DownloadingNntpClient.cs) (`DecodedBodiesPipelinedAsync` / `DecodedArticlesPipelinedAsync`).
5. **No process-wide in-flight byte budget** — connection semaphores bound concurrency, not retained bytes.
6. **UsenetSharp is not the primary buffer** — `DecodedBodiesAsync` streams through a ~1 MiB backpressured `Pipe` and reads bodies sequentially. Depth-driven GB RSS is not explained by full pre-buffering of responses 2..N inside the library.

## User-reported (unverified)

- ~220-item queue → RSS past 16 GB → exit 137; host load spiked.
- Pipelining depth 2 → 7–13 GB; depth 8 → OOM within minutes.
- Stable baseline with pipelining off: ~2–3 GB.
- Config included `usenet.article-buffer-size = 200` (5× default).

Related: #505 capped import pipelining depth at 16 (insufficient if consumer retention dominates); #477.

## Acceptance

- Streaming and queue paths share an explicit in-flight byte budget (or equivalent backpressure) so a busy host cannot retain unbounded decoded article bytes.
- `article-buffer-size` (or a successor setting) bounds **bytes** (or equivalent), not only segment count.
- Prefer pooled / non-LOH-hostile buffers where segments are materialized.
- Overview or support pack exposes a memory / in-flight gauge; throttling emits one human-friendly warning (not a stack dump).
- Focused tests cover budget enforcement under concurrent drains.

## Notes for implementers

- Separately size the WebDAV drain path vs queue `ArticleCachingNntpClient` retention — they are different amplifiers.
- Do not treat UsenetSharp pipe options as the OOM root cause; see companion UsenetSharp observability issue if a global budget needs library byte accounting.
- Request a `dotnet-counters` / GC dump from reporters before tuning numbers around the depth-2 vs depth-8 story.

hoivikaj:
## Summary
- Host-wide `usenet.in-flight-article-budget-mb` (default 512, clamp 64–8192) caps decoded article bytes retained across concurrent WebDAV streams.
- `MultiSegmentStream` leases before materializing segments; `article-buffer-size` remains the per-stream count bound.
- Support pack / overview expose leased/cap/throttle gauges; throttle logs a rate-limited Warning (no stack).

Closes #634

## Test plan
- [x] Prefetch budget tests (8 passed), including concurrent lease cap + cancel release
- [ ] rclone scrub / Arr import with raised article-buffer-size; watch gauge vs RSS


hoivikaj:
## Grok review follow-up

Reviewed against #634 while finishing the PR.

### Verified
- Process-wide `InFlightArticleBudget` with fair waiter queue + rate-limited Warning
- Lease-before-drain in `MultiSegmentStream`; `BudgetedStream` releases once
- Per-stream byte ceiling when `readBudget` is null
- DI through store factories; support pack + overview gauges; settings UI/docs
- Oversized single segment still progresses when idle (`bytes > cap && leased == 0`)

### Edit pushed before open
- Replaced a flaky concurrent `MultiSegmentStream` harness (could deadlock when producers wait on `LeaseAsync` with `CancellationToken.None` while consumers pause) with a direct concurrent `LeaseAsync` contention test that still asserts `LeasedBytes ≤ Cap` and throttle events.

Focused PrefetchBudget tests: 8 passed.

PR #672, opened 2026-07-27
fix(usenet): stop estimated segment buffers allocating twice what they need

funkypenguin:
## The problem

`DrainSegmentAsync` sizes each segment's buffer from `estimate`. When a file has recorded per-segment byte ranges that's the exact length — but otherwise it's `_estimatedSegmentSize`, i.e. `fileSize / segmentCount`, an average.

The shorter final segment drags that average **strictly below** the size of a full segment. So a full segment overflows its buffer by a few hundred bytes, and `MemoryStream` answers an overflow by doubling.

The result is exact and consistent: **every buffered segment holds 2× the memory it needs**, on the Large Object Heap, and a stream keeps `usenet.article-buffer-size` of them at once. At a ~700KB segment with the default buffer size of 40, that's **55 MiB of buffer per open read instead of 30 MiB**.

The existing comment on the field already notes the average "is off by a few bytes for every segment" — this is the consequence of that.

## The fix

Add 1/16 headroom when the size is an estimate, so a full segment fits without growing. Exact sizes are used as-is and get no padding.

For a file of `n-1` uniform segments plus a partial tail, the worst case (an empty tail) puts a full segment at `average * n / (n - 1)` — an overshoot of `1/(n-1)` — so 1/16 covers every file with 17+ segments, which is every file large enough for its buffers to matter. Smaller files may still grow once, exactly as they do today.

The division rounds up deliberately: truncating leaves the bound a byte short right at the boundary. There's a regression case for that (`InlineData(17, 720_897, 1)`).

## Side benefit for `InFlightArticleBudget`

The lease is adjusted against `buffer.Length`, while the memory actually retained is `buffer.Capacity`. A doubled buffer therefore costs roughly twice what the budget believes it does. With the doubling gone, the two agree to within the headroom — so the budget added in #651 becomes considerably more accurate for files without recorded ranges.

## Measurements

At a real segment size (716,800 bytes, average 716,716):

| | capacity |
|---|---|
| before | 1,433,432 |
| after | 761,511 |

47% less, with **no change to how many bytes are read or emitted** — only to how much is allocated to hold them.

## Testing

`dotnet test tests/NzbWebDAV.Tests` — **1176 passed, 0 failed**, including 10 new cases covering the doubling regression, the boundary case, exact sizes, unusable estimates, and `int.MaxValue` overflow.

## Context

We run a large fleet of nzbdav instances and hit this while investigating a backend that was aborting with a managed `OutOfMemoryException` under concurrent reads. #651 bounds the aggregate, which is the bigger fix; this removes the constant factor underneath it for libraries imported before per-segment ranges were recorded.

### DIFF
--- a/backend/Streams/MultiSegmentStream.cs
+++ b/backend/Streams/MultiSegmentStream.cs
@@ -748,8 +748,7 @@ public class MultiSegmentStream : FastReadOnlyNonSeekableStream
             if (lease is null)
                 lease = await LeaseSegmentBytesAsync(estimate, cancellationToken).ConfigureAwait(false);
 
-            var capacity = estimate is > 0 and <= int.MaxValue ? (int)estimate : 0;
-            var buffer = new MemoryStream(capacity);
+            var buffer = new MemoryStream(DrainBufferCapacity(estimate, hasExactSize));
             await source.CopyToAsync(buffer, cancellationToken).ConfigureAwait(false);
             var drained = buffer.Length;
             if (hasExactSize)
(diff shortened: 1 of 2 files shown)


hoivikaj:
Thanks for the analysis — the estimate-undershoot diagnosis was correct.

The retained 2× `MemoryStream` capacity issue this PR targets was addressed in #690 (`PooledBufferStream` + `ArrayPool`), and the concurrent-read OOM bound landed in #651. For the ~700KB example in the description, the pool already rents 1 MiB either way, so the advertised doubling win no longer applies on current `main`.

Most of this PR is therefore superseded / deprecated against `main` (and it currently conflicts with #690). We'll track the residual ArrayPool bucket-edge re-rent case separately in #699 and follow up there rather than rebasing this change.

PR #673, opened 2026-07-27
feat(usenet): allow a preset fraction for max-queue-connections

funkypenguin:
## The problem

`usenet.max-queue-connections` is an absolute count, clamped to the pooled connection total:

```csharp
var pool = Math.Max(1, GetUsenetProviderConfig().TotalPooledConnections);
...
return Math.Clamp(value, 1, pool);
```

So its meaning depends entirely on what a user's providers add up to, which makes it awkward to set once and leave alone — and the failure is silent:

| Pooled total | Set to 140 | Result |
|---|---|---|
| 220 | 140 | 64% of the pool — as intended |
| 100 | clamped to 100 | **the whole pool; nothing reserved** |
| 20 | clamped to 20 | **the whole pool; nothing reserved** |
| 500 | 140 | 28% — queue far more restricted than intended |

There's no error or warning in the clamped cases; the setting simply stops doing anything.

## The change

Adds `usenet.max-queue-connections-preset`, mirroring `usenet.max-download-connections-per-stream-preset` directly above it in `ConfigManager`: `low`/`medium`/`high`/`max` → `0.25`/`0.5`/`0.75`/`1.0` of the pooled budget, floored at 1 so a tiny pool still makes progress.

**Nothing changes for existing installs:**

- the absolute setting keeps precedence when set
- both unset → the historical default of letting the queue use the whole pool
- an unrecognised preset falls through to that same default, matching how the per-stream preset behaves

## Why it matters

Streaming reads already win admission at the connection pool — it's a `PrioritizedSemaphore` borrowed with `SemaphorePriority.High` while queue work uses `Low`, and the default `HighPriorityOdds = 100` means a waiting stream is always released first.

But that's admission priority, not preemption: a queue download keeps its connection until its current article finishes. So when the queue is busy enough to occupy the whole pool, a stream still waits for an article to complete before it can start. Leaving some headroom is what avoids that, and a fraction is the only way to express it once across installs with very different provider counts.

## Testing

`dotnet test tests/NzbWebDAV.Tests` — **1186 passed, 0 failed**, including 23 new cases covering each preset against both a 20- and a 220-connection pool, case-insensitivity, the floor-of-1 behaviour, unrecognised presets, and that the absolute setting still wins.

## Context

We run a large fleet of nzbdav instances whose users configure anywhere from 20 to 220 connections, which is how the clamping behaviour surfaced — one fleet-wide value is genuinely unusable today. Happy to adjust the preset names or fractions if you'd prefer different ones, or to add the UI control in a follow-up.

### DIFF
--- a/backend/Config/ConfigKeys.cs
+++ b/backend/Config/ConfigKeys.cs
@@ -43,6 +43,7 @@ public static class ConfigKeys
     public const string UsenetMaxDownloadConnectionsPerStream = "usenet.max-download-connections-per-stream";
     public const string UsenetMaxDownloadConnectionsPerStreamPreset = "usenet.max-download-connections-per-stream-preset";
     public const string UsenetMaxQueueConnections = "usenet.max-queue-connections";
+    public const string UsenetMaxQueueConnectionsPreset = "usenet.max-queue-connections-preset";
     public const string QueueWorkerCount = "queue.worker-count";
     public const string QueuePaused = "queue.paused";
     public const string QueueSpeedLimitKbps = "queue.speed-limit-kbps";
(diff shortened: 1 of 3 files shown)


hoivikaj:
Superseded by https://github.com/nzbdav/nzbdav/pull/698 (fork branch could not be pushed to; @funkypenguin's original commit is preserved unchanged as the first commit here).

@funkypenguin thanks for the contrib!

PR #692, opened 2026-07-27
fix(health): keep diagnostics faithful under write storms, circuit misses, and scrubbing

### LINKED ISSUE #680: WebDAV writes to unmatched-category items loop forever on 403 instead of backing off
# What happened?

The WebDAV backend repeatedly tries to write a metadata sidecar file into every item under an `-unmatched` category folder (a holding area for releases a *arr application couldn't match to a library entry), gets 403 Forbidden on every attempt, then immediately tries to delete the same path (404, since the write never succeeded), and moves to the next item. This repeats for every item in the backlog, forever, with no backoff and no give-up condition.

With a backlog of 100+ unmatched items this becomes a sustained loop: roughly 10 requests/sec, ~1,000 MKCOL + ~1,000 PUT + ~1,000 DELETE per 5-minute window, continuously.

Confirmed present on v0.6.4 (original nzbdav-dev codebase) and still present after upgrading to v0.9.0 specifically to test whether #651 (bounding in-flight article memory) would resolve it. It didn't — the loop is unchanged after the upgrade. That makes sense in hindsight: #651 bounds memory held during streaming/queue article buffering, a different subsystem from whatever is issuing these repeated metadata writes against the unmatched-item tree.

# Steps to reproduce

1. Configure nzbdav as the download client for a *arr application, submitting a category associated with an "unmatched" holding folder (i.e., releases the application couldn't match to a known library entry land under `completed-symlinks/<category>-unmatched/<release-name>/`).
2. Accumulate a non-trivial backlog of unmatched items over time (ours dates back roughly two months, 100+ folders).
3. Leave nzbdav running and watch the logs. The MKCOL/PUT/DELETE/403/404 cycle below repeats continuously against the entire backlog, not just one stuck item.

# Expected behavior

A write that's rejected once (403) should either not be retried at all, or retried with backoff and a give-up condition after N attempts — not repeated indefinitely, every cycle, forever, across the whole unmatched backlog.

# Technical support pack, relevant logs, or screenshots

Not using the auto-generated support pack for this one — it can still retain real file/folder names, and every path here is release-specific, so I've hand-sanitized a representative sample instead (paths/names replaced with generic placeholders, pattern and timing preserved exactly):

```text
[WRN] Cannot create directory release-name-1: forbidden
[WRN] MKCOL /completed-symlinks/example-unmatched/release-name-1/ 403 6.3 ms
[WRN] Cannot create item metadata.nfo: forbidden
[WRN] PUT /completed-symlinks/example-unmatched/release-name-1/metadata.nfo 403 2.4 ms
[WRN] DELETE /completed-symlinks/example-unmatched/release-name-1/metadata.nfo 404 5.5 ms
[WRN] Cannot create directory release-name-2: forbidden
[WRN] MKCOL /completed-symlinks/example-unmatched/release-name-2/ 403 2.4 ms
[WRN] Cannot create item metadata.nfo: forbidden
[WRN] PUT /completed-symlinks/example-unmatched/release-name-2/metadata.nfo 403 7.2 ms
[WRN] DELETE /completed-symlinks/example-unmatched/release-name-2/metadata.nfo 404 9.0 ms
```
(continues cycling through every folder in the unmatched backlog, then repeats from the top)

# Environment

OS: Linux (LXC container on Proxmox), amd64
Installation: Official Docker image, `ghcr.io/nzbdav/nzbdav:v0.9.0` (also reproduced on the original nzbdav-dev v0.6.4 build before migrating)
Integration: a *arr-family application, SABnzbd-compatible download client

# Additional context

This looks like the same general bug class as #356 (a retry-forever loop with no give-up condition, fixed in a different part of the codebase — cache directory deletion on dispose). This appears to be an unreported instance of that same pattern, in the metadata-write path against unmatched items rather than the cache-cleanup path.

Happy to grab additional diagnostics if useful, but would like to keep specific release/folder names out of anything posted here.

### LINKED ISSUE #682: A missing article resets an open provider circuit breaker, so a failing provider is never shed
## Symptom

Provider circuit breakers open and close within the same second, indefinitely. From a support pack collected 2026-07-27 (Alpine, 6 cores, Newshosting + a local NNTP proxy):

```
02:16:30 newshosting open   cooldownMs=60000
02:16:31 newshosting closed
02:16:32 newshosting open   cooldownMs=60000
02:16:32 newshosting closed
02:16:35 newshosting open   cooldownMs=60000
02:16:35 newshosting closed
```

550 transitions recorded over six days — 150 open / 150 closed for `newshosting`, 125 / 125 for the proxy — with 284 of them inside a single hour. A second, unrelated install shows the same pattern (two open/closed pairs 387 ms and 208 ms apart), so this is not deployment-specific.

## Root cause

```csharp
/// <summary>
/// Article permanently missing from retention. Counts as a miss for diagnostics
/// but does not contribute to provider-failure tripping.
/// </summary>
public void RecordArticleNotFound()
{
    Interlocked.Increment(ref _articleMissCount);
    RecordSuccess();
}
```

`RecordSuccess()` does considerably more than not-trip: it clears `_trippedUntilMs`, resets `_currentCooldown` to the initial 60 s (discarding the doubling ladder), clears `_lastFailureReason`, releases the half-open probe, and emits a `Closed` transition.

So **any article missing from retention cancels an open circuit**. A 430 is a clean server response and says nothing about provider health, yet it returns a provider that was just shed for timeouts straight to normal rotation — and satisfies the half-open probe, which is supposed to require evidence of recovery.

The doc comment is accurate about what the method must not do (contribute to tripping) and silent about what it actually does (undo a trip).

## Impact

The breaker is effectively inoperative on any install with a normal rate of missing articles, which is most of them. On the install above:

- 1,194 misses against 3,601 articles in one 5-minute bucket, roughly 4 misses per second.
- A 60-second cooldown therefore survives about 250 ms before the next miss clears it.
- `failoverReasons` still shows 97 timeouts accumulating, because the provider is never actually shed — each trip is undone before it can protect anything.

Call sites that reach this path: `ArticleBodyResult.NotFound` handling and the two `result.Success == false` branches in `MultiConnectionNntpClient`.

## Suggested fix

`RecordArticleNotFound` should count the miss and, at most, clear the failure sampling window. It must not:

- zero `_trippedUntilMs`
- reset `_currentCooldown` (that would also discard the escalation ladder for a genuinely flaky provider)
- clear `_halfOpenProbeInFlight` / `_probeStartedMs`
- emit a `Closed` transition or log "recovered"

A miss during a half-open probe is neither success nor failure. Leaving the probe slot claimed until `ProbeAbandonTimeout` is the conservative choice; releasing the slot without closing the circuit is also defensible. Worth deciding explicitly and documenting, since the current behaviour treats it as proof of recovery.

## Test gap

`ConcurrencyTests.ProviderCircuitBreaker_RecordArticleNotFoundCountsMissWithoutTripping` only asserts that a miss does not *trip* a closed breaker. Nothing asserts that a miss cannot *reset* an open one, which is why this went unnoticed. Suggested cases:

- Trip the breaker, record a miss, assert the state is still `Open` and `CooldownRemainingSeconds` is unchanged.
- Expire the cooldown into half-open, record a miss, assert the circuit does not close.
- Trip twice and assert the cooldown ladder still doubles when misses are interleaved.

### LINKED ISSUE #683: Range stall attribution bills prefetch to the wrong range after an abort
## Problem

The per-range stall attribution added in #678 credits a fetch to whichever range is current when the fetch **completes**, and `DrainForRange()` zeroes the counters at every `RangeEnd`. When a client aborts a range mid-flight, prefetch already in progress finishes afterwards and is billed to the *next* range.

The result is numbers that look plausible but are physically impossible. From a support pack with a 40-connection pool:

| Client | Median range wall clock | Median `providerWaitMs` | Implied concurrency |
|---|---|---|---|
| Browser | 226 ms | 12,580 ms | 55x |
| Plex | 591 ms | 6,114 ms | 10x |

55 concurrent fetches cannot happen on a 40-connection pool, and certainly not inside a 226 ms window. The provider time belongs to the long range that was aborted just before.

This matters because the attribution is only trustworthy for long, cleanly-terminated ranges. Scrubbing clients — the exact case the instrumentation was built for — produce short aborted ranges where the figures mislead.

## Suggested fix

Attribute at fetch **start** instead of completion:

- Assign each range a generation number, incremented on `RangeOpen` for that session.
- Capture the generation when the fetch's stopwatch starts (the same points that already call `Stopwatch.GetTimestamp()` in `MultiSegmentStream.DrainSegmentAsync`, `MultiConnectionNntpClient.AcquireConnectionLockAsync`, and `MultiProviderNntpClient.RecordFetch`).
- Add to that generation's bucket even when it resolves after the range ended, and keep a small number of recently-closed generations alive so late arrivals are not dropped.

A cheap sanity check worth emitting alongside the totals: the count of fetches attributed to the range, so a reader can spot an impossible ratio immediately.

## Also worth documenting

Trace `connWaitMs` and the connection pool's `GateWaitMs` are not comparable and were briefly misread as contradictory during triage. Trace stalls are scoped to read sessions with tracing active; pool churn counters are process-wide and cumulative, so they include queue imports and health sweeps. On a scan-heavy install the two legitimately differ by orders of magnitude (4.4 s versus 1,414 s was observed). A line in the support-pack README would prevent the next reader repeating that mistake.

## Related

- #678 introduced the attribution this corrects.

### LINKED ISSUE #684: Stuck-queue-item removals flood the warning lane, evicting every other warning
## Problem

Resolving stuck queue items logs one warning per queue record with no aggregation. A single Sonarr season pack produced **367 identical warnings in six seconds**, about 60 per second:

```
[16:34:04.382] [Warning] Resolved stuck queue item "Gintama.S01.1080p.NF.WEB-DL.DDP2.0.H.264-Kitsune" from "http://sonarr-...-svc:80" with action Remove
[16:34:04.399] [Warning] Resolved stuck queue item "Gintama.S01.1080p.NF.WEB-DL.DDP2.0.H.264-Kitsune" from "http://sonarr-...-svc:80" with action Remove
... 365 more, identical
```

In the support pack from that install, **367 of 368 warnings were this one event**. That is 74% of the 500-entry warning buffer consumed by a single occurrence; slightly more and it would have evicted the only other warning present (a cleartext-credentials notice).

One record per episode is plausible for a large season pack rather than a loop, so the removal behaviour may well be correct. The logging is not: it destroys the diagnostic value of the warning lane, which exists precisely so real warnings survive noise. #675 addressed the same class of problem for Watchtower noise crowding out streaming warnings.

## Suggested fix

Aggregate per resolution pass rather than per record — one warning naming the release, the *Arr instance, the action, and the count:

```
Resolved 367 stuck queue items for "Gintama.S01..." from sonarr (1080p) with action Remove
```

The existing helpers are the right shape to follow: `ZeroFillLogLimiter` and `ThrottledSegmentWarning` already collapse repeated per-item events into throttled summaries. Per-record detail can drop to `Log.Debug` for maintainers.

## Impact

Any install that removes a large stuck season pack loses its warning history for that window, which is exactly when an operator would be looking.

### LINKED ISSUE #685: Turning stream tracing off discards the captured trace before it can be collected
## Problem

`StreamTraceBuffer.Disable()` clears both the ring buffer and the session index:

```csharp
public StreamTraceStatus Disable()
{
    lock (_gate)
    {
        _buffer = [];
        Volatile.Write(ref _expiresAtUnixMs, 0);
        _sessions.Clear();
        Interlocked.Exchange(ref _nextSequence, 0);
    }
    ...
}
```

The natural operator flow is *enable tracing, reproduce the problem, turn tracing off, download the support pack*. That flow silently destroys the evidence. It happened during a real investigation:

```
[14:55:30] Stream tracing enabled from the UI for 30 minutes with a capacity of 20000 events
[15:00:24] Stream tracing disabled from the UI; released 1,390 buffered events
```

The pack was generated four seconds later and reported `streamTraces: disabled` with zero events. Because `WriteAsync` includes the section only when `traceStatus.Enabled || traceStatus.EventCount > 0`, and `Disable()` resets `_nextSequence` to 0 as well, there is no trace of the trace. It cost a full round trip with the reporting user to re-capture.

## Suggested fix

Separate "stop recording" from "release the memory":

- `Disable()` stops recording and lets expiry stop it, but retains the captured events so a pack collected afterwards still contains them.
- Add an explicit discard path for operators who want the RAM back immediately, or release automatically on a timer well after disabling.
- Either way `EventCount` must keep reporting what is still held, so the support pack includes the section.

If retaining is undesirable, the UI should at minimum warn that turning tracing off discards the capture and offer to download the pack first.

## Related

- #658 added the UI toggle.

hoivikaj:
## Summary
- **#682** — a missing article (430) no longer resets an open provider circuit breaker, clears its cooldown ladder, or satisfies the half-open probe
- **#680** — residual after #687: read-only write rejections throttle per mount (`completed-symlinks` shares one window) so a per-release metadata write storm collapses to one Warning (MKCOL 405 already landed in #687; the client-side write loop cannot be stopped server-side)
- **#684** — Arr stuck-queue removals log one Warning per release+action instead of one per episode record
- **#683** — range stall totals bill the range that started each fetch; late prefetch completions update the old `RangeEnd` via live generation buckets instead of the next scrub range
- **#685** — stopping stream tracing retains the capture for one hour (or until Discard) so support packs collected after turn-off still include `stream-traces/`; UI exposes retained status and an explicit discard action; re-enable resumes without wiping

## Test plan
- [ ] Backend: `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release`
- [ ] Frontend: `cd frontend && npm run typecheck && npm run build && npm test`
- [ ] Docs: `zensical build --clean --strict`
- [ ] Reproduce a 430 storm against a tripped provider — breaker stays open / half-open; no Closed transition from misses alone
- [ ] Point a metadata-writing client at completed-symlinks — one Warning per operation type per window, not one per release directory
- [ ] Configure Arr remove-on-stuck for a season pack — one Warning summarizing N removals, Debug per record
- [ ] Enable stream tracing, scrub/abort ranges under load, collect a support pack — aborted `RangeEnd` shows late `providerWaitMs`/`fetches`; next range stays clean; pair events by `rangeGeneration`
- [ ] Turn tracing off, download a support pack — `stream-traces/` included with `retained: true`; Discard clears; re-enable resumes without wiping

Fixes #680
Fixes #682
Fixes #683
Fixes #684
Fixes #685

### DIFF
--- a/backend/Api/Controllers/DiscardStreamTraces/DiscardStreamTracesController.cs
+++ b/backend/Api/Controllers/DiscardStreamTraces/DiscardStreamTracesController.cs
@@ -1,0 +1,24 @@
+using Microsoft.AspNetCore.Mvc;
+using NzbWebDAV.Api.Controllers.SetStreamTracing;
+using NzbWebDAV.Services.StreamTrace;
+using Serilog;
+
+namespace NzbWebDAV.Api.Controllers.DiscardStreamTraces;
+
+[ApiController]
+[Route("api/discard-stream-traces")]
+public sealed class DiscardStreamTracesController(
+    StreamTraceBuffer buffer,
+    StreamTraceStatusBroadcaster broadcaster) : BaseApiController
+{
+    protected override async Task<IActionResult> HandleRequest()
+    {
+        var before = buffer.GetStatus();
+        var status = buffer.Discard();
+        Log.Information(
+            "Discarded {Events:n0} retained stream-trace events from the UI",
+            before.EventCount);
+        await broadcaster.BroadcastAsync(status).ConfigureAwait(false);
+        return Ok(SetStreamTracingResponse.From(status));
+    }
+}
--- a/backend/Api/Controllers/GetStreamTraces/GetStreamTracesController.cs
+++ b/backend/Api/Controllers/GetStreamTraces/GetStreamTracesController.cs
@@ -19,8 +19,10 @@ public class GetStreamTracesController(StreamTraceBuffer buffer) : BaseApiContro
         {
             Status = true,
             Enabled = status.Enabled,
+            Retained = status.Retained,
             Source = status.Source,
             ExpiresAtUnixMs = status.ExpiresAtUnixMs,
+            RetainedUntilUnixMs = status.RetainedUntilUnixMs,
             Capacity = status.Capacity,
             EventCount = status.EventCount,
             SessionCount = status.SessionCount,
--- a/backend/Api/Controllers/GetStreamTraces/GetStreamTracesResponse.cs
+++ b/backend/Api/Controllers/GetStreamTraces/GetStreamTracesResponse.cs
@@ -5,8 +5,10 @@ namespace NzbWebDAV.Api.Controllers.GetStreamTraces;
 public class GetStreamTracesResponse : BaseApiResponse
 {
     [JsonPropertyName("enabled")] public required bool Enabled { get; init; }
+    [JsonPropertyName("retained")] public required bool Retained { get; init; }
     [JsonPropertyName("source")] public required string Source { get; init; }
     [JsonPropertyName("expiresAtUnixMs")] public required long ExpiresAtUnixMs { get; init; }
+    [JsonPropertyName("retainedUntilUnixMs")] public required long RetainedUntilUnixMs { get; init; }
     [JsonPropertyName("capacity")] public required int Capacity { get; init; }
     [JsonPropertyName("eventCount")] public required long EventCount { get; init; }
     [JsonPropertyName("sessionCount")] public required int SessionCount { get; init; }
--- a/backend/Api/Controllers/GetWebdavItem/GetWebdavItemController.cs
+++ b/backend/Api/Controllers/GetWebdavItem/GetWebdavItemController.cs
@@ -223,7 +223,8 @@ public class GetWebdavItemController(
         string? message = null)
     {
         activeReadRegistry.SetEndReason(sessionId, reason);
-        streamTrace.RangeEnd(traceRange, reason, activeReadRegistry.GetBytesRead(sessionId), message);
+        streamTrace.RangeEnd(
+            sessionId, traceRange, reason, activeReadRegistry.GetBytesRead(sessionId), message);
     }
 
     private async Task CopyAndReportAsync(
--- a/backend/Api/Controllers/GetWebdavItem/GetWebdavItemController.cs
+++ b/backend/Api/Controllers/GetWebdavItem/GetWebdavItemController.cs
@@ -163,6 +163,7 @@ public class GetWebdavItemController(
             readCts.CancelAfter(configManager.GetStreamingReadTimeout());
             var ct = readCts.Token;
 
+            StreamTraceRangeContext? traceRange = null;
             try
             {
                 await using var response = await GetWebdavItem(request, ct);
@@ -170,7 +171,7 @@ public class GetWebdavItemController(
                     return;
                 var effectiveStart = (long)(HttpContext.Items["effectiveRangeStart"] ?? 0L);
                 var rangeEnd = HttpContext.Items["effectiveRangeEnd"] as long?;
-                streamTrace.RangeOpen(
+                traceRange = streamTrace.RangeOpen(
                     sessionId,
                     request.Item,
                     "GET",
@@ -179,28 +180,29 @@ public class GetWebdavItemController(
                     response.CanSeek ? response.Length : null,
                     Request.Headers.UserAgent.ToString(),
                     HttpContext.Connection.RemoteIpAddress?.ToString());
+                using var traceRangeScope = MultiProviderNntpClient.BeginStreamTraceRangeScope(traceRange);
                 try
                 {
                     // Body transfer can run for minutes; drop the admission/open
                     // deadline and rely on per-segment mid-stream timeouts.
                     readCts.CancelAfter(Timeout.InfiniteTimeSpan);
-                    await CopyAndReportAsync(response, Response.Body, sessionId, effectiveStart, ct);
-                    FinishRange(sessionId, ReadSession.EndReasonCode.Completed);
+                    await CopyAndReportAsync(response, Response.Body, sessionId, effectiveStart, traceRange, ct);
+                    FinishRange(sessionId, traceRange, ReadSession.EndReasonCode.Completed);
                 }
                 catch (OperationCanceledException) when (HttpContext.RequestAborted.IsCancellationRequested)
                 {
-                    FinishRange(sessionId, ReadSession.EndReasonCode.Aborted);
+                    FinishRange(sessionId, traceRange, ReadSession.EndReasonCode.Aborted);
                     throw;
                 }
                 catch (Exception ex)
                 {
-                    FinishRange(sessionId, ReadSession.EndReasonCode.Error, ex.Message);
+                    FinishRange(sessionId, traceRange, ReadSession.EndReasonCode.Error, ex.Message);
                     throw;
                 }
             }
             catch (OperationCanceledException oce) when (!HttpContext.RequestAborted.IsCancellationRequested)
             {
-                FinishRange(sessionId, ReadSession.EndReasonCode.Error, "streaming-read-timeout");
+                FinishRange(sessionId, traceRange, ReadSession.EndReasonCode.Error, "streaming-read-timeout");
                 throw new StreamingReadTimeoutException(
                     "WebDAV /view read exceeded the " +
                     $"{configManager.GetStreamingReadTimeout().TotalSeconds:0}s streaming-read-timeout " +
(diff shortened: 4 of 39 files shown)


hoivikaj:
## Product review (post-fix)

**Verdict:** Approve with the follow-up commit on the branch (`954ce1f6`).

### Claim validation
- **#682** → confirmed — `RecordArticleNotFound` no longer calls `RecordSuccess`; open/half-open/ladder covered.
- **#680** → partial / completes residual after #687 — MKCOL 405 already on `main`; this PR only coarsens throttle scope to the mount (`completed-symlinks`). Client-side write loop remains client-side (as diagnosed on the issue). PR body clarified.
- **#684** → confirmed — one Warning per title+action; Debug per record.
- **#683** → confirmed — generation tokens + live `RangeStalls`; AsyncLocal reaches prefetch via lazy `NzbFileStream` creation; null-range `RangeEnd` still records.
- **#685** → confirmed — `StopRecording` retains; `Discard` clears; resume path tested. Follow-up: `EnableFor` while already recording now refreshes TTL instead of wiping.

### Corrections landed
1. **CI flake** — `LogResolutionSummary_EmitsOneWarningPerGroup` asserted on every Warning in the process logger; parallel `ProviderCircuitBreaker` trip Warnings made it fail. Now filters on `QueueItemTitle`.
2. **`EnableFor` while recording** — no longer wipes an in-progress capture (UI enable over `STREAM_TRACE_EVENTS`, or accidental re-enable).

### Notes
- Tests: adequate after the flake fix.
- Breaking / migration: no.
- Residual risk: #680 client still retries writes; operators stop that client-side. Trace retention holds RAM for up to 1h after stop unless Discarded.

PR #697, opened 2026-07-28
fix(usenet): prefer the configured primary over larger idle backup pools

hoivikaj:
## Summary
- Cascade was ranking providers by absolute idle connection count, so a Priority-0 primary with fewer MaxConnections lost to larger lower-priority pools even while idle (seen in support packs where Frugal Bonus served ~85% of fetches while Newshosting was ~8× faster in the small observed sample).
- Cascade now preserves configured priority while more than 25% of a provider pool remains free, then lets an idle next-priority peer take work when the primary becomes thinly spared.
- Equal-utilization cascade ties use spare fraction rather than absolute connection count, preventing larger pools from regaining the same bias in the later tie-break.

## Test plan
- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --filter FullyQualifiedName~MultiProviderNntpClient`
- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release`
- [ ] After merge: ask the original reporter to restore all-Pool Connections briefly, re-collect a Plex stream trace, and confirm Newshosting receives most successful fetches while it has healthy spare capacity

## Reporter interim guidance
Keep cascade on. If Newshosting is the intended primary, mark true fallbacks as **Backup Only** until this ships — do not inflate MaxConnections or raise article-buffer-size as the first workaround.

### DIFF
--- a/backend/Clients/Usenet/Connections/ConnectionPool.cs
+++ b/backend/Clients/Usenet/Connections/ConnectionPool.cs
@@ -46,6 +46,7 @@ public sealed class ConnectionPool<T> : IDisposable, IAsyncDisposable
     private const int MaxConcurrentHandshakes = 3;
 
     public TimeSpan IdleTimeout { get; }
+    public int MaxConnections => _maxConnections;
     public int LiveConnections => _live;
     public int IdleConnections => _idleConnections.Count;
     public int ActiveConnections => _live - _idleConnections.Count;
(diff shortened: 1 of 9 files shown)


hoivikaj:
## Review follow-up

Applied on this branch in `8f0d7317`:

1. Use-cases doc no longer promises “strict priority” (thin-spare yield still applies).
2. Thin-spare check uses integer `unreserved * 4 <= max` so 2/8 does not depend on float.

Cascade-focused tests: 8/8 passed. Remaining notes are non-blocking (squash the add/revert corrupt-retry commits; keep corrupt-retry + analyzer pairing as separate follow-ups).

Issue #704, opened 2026-07-28
Background health checks can outcompete playback despite streaming priority


hoivikaj:
## Problem

Background health checks and playback use independent connection budgets. The Streaming Priority setting does not arbitrate between them, so maintenance can receive substantially more provider connections than an actively starving stream.

## Expected

When playback is active—especially when trace data shows the consumer waiting for prefetch—streaming priority should reserve or reclaim enough shared provider capacity to keep playback smooth. Background checks should yield or be capped accordingly.

## Observed

A streaming support pack used:

- `repair.healthcheck-concurrency=32`
- Queue connections: Auto (all configured provider connections)
- `usenet.streaming-priority=95`
- `article-buffer-size=32` with four-article BODY batches, giving playback an effective width of roughly eight connections

Health-check STAT traffic was active during the test, including 526 article misses in the 13:50 and 13:55 metrics buckets. During the same period, a stream spent 80–97% of range time waiting for prefetch after provider latency rose, while most configured capacity remained available.

## Environment

- Version: `dev-dab9146`
- Reporter: community streaming-test report
- Setup: NzbDAV and rclone in Docker on an Oracle Cloud ARM64 VPS; background repairs enabled

hoivikaj:
### Maintainer notes

**Likely cause:** `DownloadingNntpClient` creates separate `_streamingSemaphore` and `_queueSemaphore` budgets. Health-check STAT requests use the queue semaphore; WebDAV reads use the streaming semaphore. `usenet.streaming-priority` configures odds only within the streaming semaphore, so it cannot make health/queue work yield. Both paths meet only at the per-provider connection-pool gate.

**Product vs config:** Product bug with a configuration workaround. Reducing health-check concurrency or scheduling checks outside viewing hours helps, but the advertised streaming-priority control does not govern the competing maintenance work operators expect it to prioritize against.

**Evidence:** This host allowed 32 concurrent health checks while the active stream's fixed prefetch geometry used roughly eight connections. A scan was active during the tested window, and the stream simultaneously showed sustained consumer starvation. Pool capacity was not globally exhausted, so static independent budgets—not a lack of configured connections—prevented useful reallocation.

[...]

PR #707, opened 2026-07-28
fix(streaming): prioritize playback, adapt prefetch, and improve support diagnostics

### LINKED ISSUE #699: Avoid ArrayPool bucket-edge re-rents when draining estimated segments
## Problem

When `MultiSegmentStream` drains a segment without a recorded exact size, the capacity hint is `fileSize / segmentCount` (an average that undershoots full segments). After #690, drains use `PooledBufferStream` over `ArrayPool<byte>.Shared`. That removed the old `MemoryStream` 2× doubling, but a small undershoot can still straddle an ArrayPool bucket boundary: the initial rent fits the estimate, then `EnsureCapacity` re-rents the next bucket and copies once.

## Expected

Capacity hints for estimated segments should usually land in the same pool bucket as the decoded full segment, avoiding a grow/copy on the common path. Any padding must not push mid-bucket hints into a larger bucket when pool slack already covers the undershoot.

## Observed

For typical ~700KB segments, `Rent(~716716)` already returns 1 MiB, so the undershoot does not grow (the #672 scenario is already fine). Near power-of-two bucket edges (e.g. ~512 KiB), estimate vs full size can still force a re-rent. Naive 1/16 headroom (as in #672) can also *increase* retained size by jumping the next bucket when slack would have been enough.

## Context

- Supersedes the residual concern from #672 after #690 landed.
- OOM under concurrent reads remains bounded by #651 (`usenet.in-flight-article-budget-mb`).
- Candidate area: capacity hint passed into `PooledBufferStream` from `DrainSegmentAsync` in `backend/Streams/MultiSegmentStream.cs`, and/or grow policy in `backend/Streams/PooledBufferStream.cs`.

### LINKED ISSUE #703: Playback can starve during provider latency spikes while download connections sit idle
## Problem

Playback throughput can collapse during a provider latency spike even when most configured download connections are idle. NzbDAV uses a fixed per-stream prefetch width, so it does not widen the pipeline when the consumer starts starving.

## Expected

When article latency rises and the player is waiting for prefetched data, NzbDAV should use available streaming capacity to preserve playback throughput, up to the configured download and memory limits.

## Observed

A streaming support pack captured one ten-minute starvation window:

- Newshosting article latency rose from a 52 ms mean / 213 ms max to a 348 ms mean / 4,987 ms max.
- 7.2% of articles took over one second, but all 3,604 traced segments were `Ok` with zero retries.
- Range attribution showed 80–97% `consumerWaitMs` while the player waited for prefetch.
- `article-buffer-size=32` with four-article BODY batches limited the stream to roughly eight connections, despite hundreds of provider connections being available.
- The predicted ceiling was 16.4 MB/s; the observed ceiling was 16.9 MB/s.

Other retained ranges were client-paced rather than server-limited, reaching up to 82 MB/s. The first three titles cannot be evaluated from segment events because the trace buffer had already overwritten them; see #705.

## Environment

- Version: `dev-dab9146`
- Reporter: community streaming-test report
- Setup: NzbDAV and rclone in Docker on an Oracle Cloud ARM64 VPS; Plex playback through an rclone VFS mount

### LINKED ISSUE #705: Stream tracing silently discards most of long playback tests when its buffer fills
## Problem

The stream-trace ring buffer silently overwrites old events when it reaches capacity. Support packs do not indicate that the capture is incomplete, so maintainers can mistake the retained tail for the full test and miss the reported failure window.

## Expected

The Support UI and support pack should clearly report trace overflow and the retained time window. A normal multi-title playback test should either fit by default or guide the operator to increase capacity before reproducing the problem.

## Observed

A 25-minute, six-title streaming test produced:

- `eventCount: 48040`
- `capacity: 20000`
- only 20,000 events in `stream-traces/events.jsonl`
- trace events spanning only 14:00:17–14:10:27
- 13 sessions listed in `sessions.json`, but retained events for only five sessions

Approximately 58% of recorded events were evicted. The first three titles were entirely absent from the retained segment events and the fourth title was already in progress when the retained window began. This includes the earlier interval containing all article-miss metrics. The pack and UI gave no overflow warning.

## Environment

- Version: `dev-dab9146`
- Reporter: community streaming-test report
- Capture: UI stream tracing, default 20,000-event capacity, support pack collected with tracing left enabled

### LINKED ISSUE #706: Add per-provider latency percentiles over time to support packs
## Problem

Support packs show provider throughput, misses, failures, and circuit transitions over time, but not provider latency percentiles. When playback slows without errors, maintainers cannot tell whether the provider became slow, local connection contention increased, or the streaming pipeline under-filled.

## Expected

Support packs should include a bounded per-provider latency time series with enough separation to distinguish successful provider response latency from local admission/pool waits.

## Observed

In a streaming test, trace events showed one provider's segment duration change from:

- mean 52 ms / max 213 ms during a healthy title
- mean 348 ms / max 4,987 ms during a starvation window
- 7.2% of segments over one second in the bad window

All segments were `Ok` with zero retries. `metrics/recent.json` preserved five-minute throughput and seven-day provider consumption, but no latency distribution over those intervals. The trace established correlation for the retained tail only; it could not prove whether the change originated at the provider or in local contention, and older trace events had already been overwritten.

## Environment

- Version: `dev-dab9146`
- Reporter: community streaming-test report
- Setup: multiple NNTP providers, NzbDAV/rclone/Plex streaming test

## Related

- #259 requests live per-provider sustained MB/s; this issue is specifically historical latency diagnosis in support packs.
- #637 addressed misleading/conflated provider latency and contention behavior.

---

hoivikaj:
## Summary
- **#705** — Stream-trace UI/support packs report retained vs overwritten events, overflow warnings, selectable capacity (default 100k), and stream JSONL into packs (`schemaVersion` 3 + `OVERFLOW.txt`).
- **#704** (partial) — Streaming Priority odds now apply at each provider connection-pool gate under contention (Profile Play verification is High). Spare capacity is still not reserved for playback — health STAT can fill free slots on an unsaturated pool; follow-up remains for that residual.
- **#706** — Named latency phases (`response`, `pool-wait`, `permit-wait`) are histogrammed into `MetricEvent` rows and projected as support-pack `latency24Hours` (no EF migration).
- **#703** — BODY pipeline batch width adapts from consumer readiness (4→2→1 when starved; gradual recovery) so high latency can use more connections at the same article-buffer memory cost.
- **#699** — Drain capacity hints prefer plausible yEnc `PartSize` over the file average to avoid ArrayPool grow-and-copy near bucket edges.

### Operator note: Streaming Priority default at the provider gate
On previous builds the provider connection-pool gate used implicit High odds of **100** (playback won every contested release). This PR wires the configured Streaming Priority setting into that gate; the setting’s default is **80**, so at default settings Low (queue/health) wins roughly one in five contested releases under saturation. Set Streaming Priority to 100 to keep the old “always yield to playback when contested” behavior. Docs already state that spare (unsaturated) capacity is never held idle for High.

Fixes #705
Fixes #706
Fixes #703
Fixes #699
Relates to #704

## Test plan
- [x] `dotnet build backend/NzbWebDAV.csproj -c Release`
- [x] Full backend suite (`dotnet test tests/NzbWebDAV.Tests/… -c Release`) — 1354 passed, ×25 under parallel load to confirm the adaptive-width tests are load-stable
- [x] Frontend typecheck + `backend-client.server.test.ts`
- [ ] Manual: enable stream tracing at 20k, overflow a long playback, confirm UI banner + pack `OVERFLOW.txt`
- [ ] Manual: saturate provider with health concurrency + playback; confirm Streaming Priority shares connections
- [ ] Manual: induce provider latency; confirm adaptive batch narrows and throughput recovers
- [ ] Manual: support pack `metrics/recent.json → latency24Hours` shows phase series after load

### Adaptive width (#703): what is covered where
`AdaptiveBodyBatchSizer` owns the 4→2→1 narrowing and the 1→2→4 recovery ladder, and those transitions are asserted deterministically in `AdaptiveBodyBatchSizerTests`. The `MultiSegmentStream` integration tests assert the wiring around it: that real per-boundary readiness samples reach the sizer, that narrowing happens end to end under starvation, that FIFO order and byte-exactness survive width transitions in both directions, and that batch widths stay inside the configured band.

Sustained end-to-end *widening* is deliberately not asserted: it needs 32 consecutive ready boundaries, which against an in-memory producer depends on scheduling rather than on the code under test, and flaked roughly a third of the time under parallel suite load. The manual latency check above is what confirms recovery on a real provider.

### DIFF
--- a/backend/Api/Controllers/ClearOverviewStats/ClearOverviewStatsController.cs
+++ b/backend/Api/Controllers/ClearOverviewStats/ClearOverviewStatsController.cs
@@ -11,7 +11,8 @@ namespace NzbWebDAV.Api.Controllers.ClearOverviewStats;
 public class ClearOverviewStatsController(
     ConfigManager configManager,
     MetricsWriter metricsWriter,
-    ProviderBytesTracker bytesTracker
+    ProviderBytesTracker bytesTracker,
+    ProviderLatencyTracker latencyTracker
 ) : BaseApiController
 {
     protected override async Task<IActionResult> HandleRequest()
@@ -34,38 +35,40 @@ public class ClearOverviewStatsController(
 
         // 2. Pause flushes and abandon any in-flight drained batch, then drop
         //    queued rows so they cannot reappear after the wipe.
-        metricsWriter.BeginReset();
-        try
+        await using var resetLease = await metricsWriter.BeginResetAsync(ct).ConfigureAwait(false);
+
+        if (providerKey == null) metricsWriter.DiscardQueuedAndResetStats();
+        else metricsWriter.DiscardQueuedForProvider(providerKey);
+
+        // 3. Wipe metrics tables (all, or provider-keyed rows only).
+        await using var db = new MetricsDbContext();
+        var deletedRows = providerKey == null
+            ? await OverviewStatsReset.WipeAsync(db, ct).ConfigureAwait(false)
+            : await OverviewStatsReset.WipeProviderAsync(db, providerKey, ct).ConfigureAwait(false);
+
+        // 4. Zero in-memory lifetime counters and pending minute buckets.
+        if (providerKey == null)
         {
-            if (providerKey == null) metricsWriter.DiscardQueuedAndResetStats();
-            else metricsWriter.DiscardQueuedForProvider(providerKey);
-
-            // 3. Wipe metrics tables (all, or provider-keyed rows only).
-            await using var db = new MetricsDbContext();
-            var deletedRows = providerKey == null
-                ? await OverviewStatsReset.WipeAsync(db, ct).ConfigureAwait(false)
-                : await OverviewStatsReset.WipeProviderAsync(db, providerKey, ct).ConfigureAwait(false);
-
-            // 4. Zero in-memory lifetime counters and pending minute buckets.
-            if (providerKey == null) bytesTracker.ResetCounters();
-            else bytesTracker.ResetProvider(providerKey);
-
-            // 5. Fold the pre-wipe usage snapshot into BytesUsedOffset and
-            //    persist. Done after the wipe so SeedTrackerAsync cannot read
-            //    stale ProviderHourly rows or double-count live tracker bytes.
-            if (OverviewStatsReset.FoldUsageIntoOffsets(providerConfig, usageSnapshot, nowMs, providerKey))
-                await UsenetProviderIdentity.SaveProvidersAsync(configManager, providerConfig, ct)
-                    .ConfigureAwait(false);
-
-            // 6. Drop anything enqueued during the wipe window.
-            if (providerKey == null) metricsWriter.DiscardQueuedAndResetStats();
-            else metricsWriter.DiscardQueuedForProvider(providerKey);
-
-            return Ok(new ClearOverviewStatsResponse { Status = true, DeletedRows = deletedRows });
+            bytesTracker.ResetCounters();
+            latencyTracker.ResetCounters();
         }
-        finally
+        else
         {
-            metricsWriter.EndReset();
+            bytesTracker.ResetProvider(providerKey);
+            latencyTracker.ResetProvider(providerKey);
         }
+
+        // 5. Fold the pre-wipe usage snapshot into BytesUsedOffset and
+        //    persist. Done after the wipe so SeedTrackerAsync cannot read
+        //    stale ProviderHourly rows or double-count live tracker bytes.
+        if (OverviewStatsReset.FoldUsageIntoOffsets(providerConfig, usageSnapshot, nowMs, providerKey))
+            await UsenetProviderIdentity.SaveProvidersAsync(configManager, providerConfig, ct)
+                .ConfigureAwait(false);
+
+        // 6. Drop anything enqueued during the wipe window.
+        if (providerKey == null) metricsWriter.DiscardQueuedAndResetStats();
+        else metricsWriter.DiscardQueuedForProvider(providerKey);
+
+        return Ok(new ClearOverviewStatsResponse { Status = true, DeletedRows = deletedRows });
     }
 }
--- a/backend/Api/Controllers/GetStreamTraces/GetStreamTracesController.cs
+++ b/backend/Api/Controllers/GetStreamTraces/GetStreamTracesController.cs
@@ -26,6 +26,13 @@ public class GetStreamTracesController(StreamTraceBuffer buffer) : BaseApiContro
             Capacity = status.Capacity,
             EventCount = status.EventCount,
             SessionCount = status.SessionCount,
+            RetainedEventCount = status.RetainedEventCount,
+            OverwrittenEventCount = status.OverwrittenEventCount,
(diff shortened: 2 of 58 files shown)


hoivikaj:
Follow-up from review landed in 765d8b7e:

1. **CI / adaptive-width flake** — `TestOnSegmentReadiness` is now instance-scoped on `MultiSegmentStream` so parallel stream tests cannot trip another test’s lockstep TCS. Full suite green locally (1354 passed ×5).
2. **Header-throw drain fallback** — added theory covering `InvalidDataException` / `IOException` from `GetYencHeadersAsync`.
3. **PR narrative** — disclosed the provider-gate default odds change (implicit 100 → configured default 80); marked **#704** as partial and switched to `Relates to #704` so it does not auto-close (unsaturated health STAT residual remains).

hoivikaj:
Follow-up review of my own previous flake fix — it made the adaptive-width tests green by weakening what they test, so I've reverted that approach.

**What was wrong**

1. The previous fix set `TestSuppressReadinessObserve` to disable the production readiness observation and then fed the sizer a hardcoded `readyWhenNeeded: false`. The integration tests therefore no longer exercised the readiness path they exist to cover — they had become a slower duplicate of `AdaptiveBodyBatchSizerTests`. Both seams are now removed from `MultiSegmentStream`, which is back to a single hook invocation and no extra hot-path branch.

2. `Recovery_ReturnsGraduallyToFourAfterSustainedReadiness` asserted `Contains(2)` and `Contains(4)` against `ObservedBatchSizes`, which were already satisfied by the width-4 batches issued *before* the pipeline narrowed. The assertion could not fail for the reason the test existed. Reading the sizer instead exposed that the width never recovered at all in that harness.

**Why the starvation flake actually happened**

The producer reads `_batchSizer.Current` once per loop iteration and then blocks writing that batch into a full channel. While it is blocked the consumer can cross enough boundaries to narrow 4→2→1, so the producer resumes and issues at 1 without ever emitting a width-2 batch. Asserting on issued batch sizes was the bug; width is now asserted from the sizer, with real per-boundary samples flowing through the production path (`Assert.All(readinessSamples, ready => Assert.False(ready))`).

**Recovery coverage**

Sustaining 32 consecutive ready boundaries against an in-memory producer is scheduling-bound, not code-bound — it failed 7/20 full-suite runs. `AdaptiveBodyBatchSizerTests.SixteenConsecutiveReady_RecoversOneStepAtATime` already covers the 1→2→4 ladder deterministically, so the integration test now covers the load-stable part (ready boundaries resume, widths stay in band, bytes stay exact across the narrow region) and is named `ReadyBoundariesResume_AfterStarvationNarrowsPipeline`. The coverage boundary is written into the PR body so it isn't mistaken for end-to-end recovery proof.

Also silenced the `CS4014` warnings this PR introduced in `ConnectionPoolPriorityTests`.

**Verification:** 31 full-suite runs (1354 tests) — 30 clean; the one failure was an unrelated DB task test in the run immediately after I killed a hung stress loop, and it passes 10/10 in isolation. CI is green on `13479858`.

Issue #461, opened 2026-07-18
Classify missing video segments as fatal vs degraded


hoivikaj:
### Before submitting

- [x] I have searched existing issues and did not find a similar request.
- [x] This request relates to use with legally obtained content.

### Problem or use case

Any missing Usenet segment is treated as binary corruption: urgent health, repair, and often Arr redownload. For MP4/MKV libraries, a missing mid-stream media payload range may only cause a short glitch that players skip, while a missing `moov`/EBML header/cluster index can make the file unplayable.

Operators need health/repair to distinguish **fatal** (unplayable) from **degraded** (still playable) so Arr is not churned for skippable gaps.

Related but different: [#257](https://github.com/nzbdav/nzbdav/issues/257) skips non-video/audio entirely; this issue is about impact classification **within** video files.

### Proposed solution

- Probe container structure (MP4 boxes / MKV EBML) once after import (or lazily with a bound), persist a compact media-structure map on the item.
- When a missing segment (or streaming failure offset) is known, classify offline:
  - **fatal** — overlaps container-critical structure → today’s repair path
  - **degraded** — media payload (or seek-index-only) → record health status, **do not** trigger Arr repair
  - **unknown** — non-video, encrypted, unparseable → treat as fatal (safe default)
- Surface degraded status in Health UI so operators can still see soft failures.

### Alternatives considered

- Acceptable missing-segment percentage only — does not know *where* the gap is.
- Always repair — current behavior; maximizes Arr churn.

### Relevant area

Queue, NZB processing, or health checks

### Additional context

Came out of a review of health/repair workflow issues for stream-first libraries where players tolerate mid-file payload holes better than Arr repair loops do.

---

**Upvote this request** by reacting with a thumbs-up (👍) on this issue. Thumbs-up counts are tracked as Votes on the project board and help us prioritize feature work.

hoivikaj:
## Triage note

Lowered **Priority → Medium**. Fatal-vs-degraded segment classification should follow #460 (threshold before Arr repair); larger media-structure design than the first repair-gating step.

loambit:
I was looking to implement this one myself to reduce the amount of churn in my library until I saw you already had an issue lined up. Claude helped with the design on this one:

## Approach

Classify by how much is missing rather than by where it lands, which defers the container probe rather than replacing it. Sequenced as you described on the triage note, after the repair-gating step.

## Flow

1. The scheduled health check sweeps and counts instead of aborting on the first miss.
[...]

PR #709, opened 2026-07-28
feat(health): tolerate a few missing articles in plain video files

loambit:
First pass at issue #461, an optional setting to enable more lenient health checks for certain safe filetypes. Allows up to 64 articles or 2% (whichever is smaller) missing before marking it a failed health check. Seems to significantly reduce the amount of churn in my library for non multipart releases, especially for certain shows, and I haven't noticed a difference in playback. 
Archives and encrypted files are excluded, since a zero filled gap there breaks the extraction or decodes to noise rather than causing a brief glitch. Three consecutive missing articles fails the file regardless of the totals, using the same limit the read path uses before it gives up zero filling, so the health check can't pass something playback would fail.
Future changes would include some UI labels to indicate degraded files, re-checking degraded files sooner than clean ones so damage that grows gets caught earlier, and improving checks to detect missing keyframes.

Screenshot 2026-07-28 163702

### DIFF
--- a/backend/Clients/Usenet/ArticleExistenceChecker.cs
+++ b/backend/Clients/Usenet/ArticleExistenceChecker.cs
@@ -6,11 +6,12 @@ internal static class ArticleExistenceChecker
     /// Checks health/import existence concurrently across the connection pool.
     /// BODY pipelining settings must not collapse these probes onto one connection.
     /// </summary>
-    public static Task CheckAsync(
+    public static Task<IReadOnlyList<int>> CheckAsync(
         INntpClient client,
         IReadOnlyList<string> segmentIds,
         int concurrency,
         IProgress<int>? progress,
-        CancellationToken cancellationToken) =>
-        client.CheckAllSegmentsAsync(segmentIds, concurrency, progress, cancellationToken);
+        CancellationToken cancellationToken,
+        int toleratedMissing = 0) =>
+        client.CheckAllSegmentsAsync(segmentIds, concurrency, progress, cancellationToken, toleratedMissing);
 }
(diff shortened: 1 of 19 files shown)


hoivikaj:
@loambit Thanks for putting together this first pass. I reviewed the implementation against issue #461 and the current `main` branch.

## Review verdict: Request changes

### Must-fix

1. **Degraded checks are reported as Healthy.**
   `frontend/app/routes/health/components/health-stats/health-stats.tsx` counts every `HealthResult.Healthy`, including `(Healthy, Degraded)`. The new `Degraded` enum value is currently unused. Please add a degraded category and exclude degraded checks from healthy totals.

2. **Sampled checks cannot safely enforce the advertised limits.**
   `HealthCheckService` calculates tolerance from the whole file but applies it only to the sampled articles. A partial check can therefore accept a file with more than 64/2% missing articles, or miss a three-article run that playback will reject. Please require complete coverage before returning degraded, or conservatively repair when a partial sample finds damage.

3. **The documentation references the wrong release.**
   This feature was not included in released `v0.9.2`. Both `since 0.9.2` pills should target the release that actually introduces it, currently expected to be `0.9.3`.

### Completeness and tests

- Issue #461 is only partially addressed: degraded results are persisted, but operators cannot identify degraded files in the Health UI.
- Please add coverage for degraded stats bucketing and the end-to-end health-result classification.
- The branch is two commits behind `main` and overlaps recent health work. It is currently mergeable, but checks should rerun after updating it.

### Confirmed

- The setting is off by default and fully wired.
- Archive and multipart members are conservatively excluded.
- The shared three-consecutive-miss threshold matches normal buffered playback.
- Current CI passes, and no migration or breaking marker is required.

loambit:
The code review agent wants a bigger implementation and there's been lots of refactoring in the same are of the codebase so I'll close this and keep it on my own fork for now.

PR #714, opened 2026-07-29
fix(metrics): Overview Backup rescues no longer counts same-provider retries

### LINKED ISSUE #713: Overview Backup rescues counts same-provider retries as failover saves
## Problem

The Overview **Backup rescues** card counts same-provider retries (for example Newshosting timing out, then succeeding on the next attempt) as if a backup provider rescued the segment. Single-provider setups still show rescues even with no Backup Only account configured.

## Expected

Backup rescues should only count when a **different** provider successfully serves a segment after another provider missed or failed. Same-host self-retries after timeouts or transient errors should not inflate the Backup rescues card.

## Observed

With a single Newshosting provider and no backup:

- Backup rescues shows non-zero segments rescued
- **Most saves** and **Most misses** are both `news.newshosting.com`
- **Why they missed** is entirely `timed out`

That pattern is a same-provider retry, not a backup stepping in.

## Environment

- Version: 0.9.1 / local main
- Reporter: maintainer (hoivikaj)
- Setup notes: single pooled provider, no Backup Only provider

## Logs

N/A — reproduced from Overview UI metrics (same host on both sides of the rescue edge).

hoivikaj:
## Summary
- Fixes #713: same-provider self-retries (e.g. Newshosting timeout → retry succeeds) no longer inflate Overview **Backup rescues**
- Failover save / miss metrics and Ok `Retries` for rollup only include cross-provider edges
- Clarifies Backup rescues card copy to say “another provider”

## Test plan
- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --filter FullyQualifiedName~MultiProviderNntpClientTests`
- [ ] Single-provider setup: play something that used to show same-host timeout rescues; new window should stay empty (or only true multi-provider failover)
- [ ] Two providers: force a miss on primary and success on backup; Backup rescues still increments
- [ ] Note: historical metrics already written keep old same-provider counts until retention ages them out

### DIFF
--- a/backend/Clients/Usenet/MultiProviderNntpClient.cs
+++ b/backend/Clients/Usenet/MultiProviderNntpClient.cs
@@ -407,12 +407,14 @@ public class MultiProviderNntpClient(
                         if (responseType == UsenetResponseType.ArticleRetrievedBodyFollows)
                         {
                             _usageTracker.RecordSuccess(provider.MetricsKey);
+                            var crossMisses = FilterCrossProviderMisses(
+                                priorMisses, provider.MetricsKey);
                             RecordFetch(provider.MetricsKey, SegmentFetch.FetchStatus.Ok,
-                                stopwatch.ElapsedMilliseconds, priorMisses?.Count ?? 0, traceRange);
-                            if (priorMisses is { Count: > 0 })
+                                stopwatch.ElapsedMilliseconds, crossMisses?.Count ?? 0, traceRange);
+                            if (crossMisses is { Count: > 0 })
                             {
                                 _usageTracker.RecordFailoverSave();
-                                RecordFailoverMisses(priorMisses, provider.MetricsKey);
+                                RecordFailoverMisses(crossMisses, provider.MetricsKey);
                             }
                             response = WrapProviderResponse(response, provider.MetricsKey);
                             gateOwnedByTransfer = true;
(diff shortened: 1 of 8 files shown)


hoivikaj:
## Follow-up from review

Pushed `86339d43` to address the Retries-column side effect called out in review.

### What changed
- **Ok-row `Retries`** again records the true attempt count (`priorMisses.Count` / `attemptIndex`), so the provider scoreboard Retries column and support-pack export still show same-provider self-retries (e.g. timeout → success on the same host).
- **`ProviderMinutes.FailoverSaves`** is now derived from distinct `FailoverMisses.At` values for that rescuer, not from `Status = 0 AND Retries > 0`. Backup rescues stay cross-provider-only.
- **Stream traces** still emit failover edges for *all* prior misses (including same-provider retries) for support-pack stall attribution; only the metrics edge table gets the filtered list.
- Added rollup tests covering same-provider retry (Retries preserved, FailoverSaves = 0), cross-provider rescue (FailoverSaves = 1), and multi-edge same-`At` collapsing to one save.

### Still intentionally out of scope
Two accounts on the same backbone remain distinct `MetricsKey`s and still count as cross-provider rescues. Worth a follow-up if we want host/storage-group dedupe.

### Checklist from original review
| Item | Status |
|------|--------|
| Same-provider retries excluded from Backup rescues | still fixed |
| Scoreboard Retries not silently zeroed | fixed in this commit |
| Stream-trace same-host edges preserved | fixed in this commit |
| Rollup tests for the headline metric | added |

PR #740, opened 2026-07-31
fix(auth)!: prevent concurrent onboarding from creating multiple admins

WhispersOfJ:
## Summary

`CreateAccountController` had no server-side awareness of whether an admin account already existed. The "only one admin, only during onboarding" invariant was enforced solely by a check-then-act in the frontend (`isOnboarding()` then `createAccount()`), a different process than the actual DB write in the backend. Two concurrent onboarding submissions during the setup window (e.g. an attacker reaching a freshly deployed, not-yet-configured instance at the same time as the legitimate operator) could both pass that frontend check and both successfully insert an `Admin` row.

- `backend/Database/DavDatabaseContext.cs` — adds a filtered unique index (`IX_Accounts_SingleAdmin`) on `Accounts.Type` where `Type = Admin`, as the authoritative DB-level backstop.
- `backend/Database/Migrations/20260731171110_Add-SingleAdmin-UniqueIndex.cs` (+ Designer + snapshot) — the corresponding EF migration.
- `backend/Api/Controllers/CreateAccount/CreateAccountController.cs` — adds an application-level pre-check for the fast/common-case rejection (clean `400` instead of a raw DB exception), plus a `catch (DbUpdateException)` for when two requests still race past the pre-check — following the existing `IsCategoryFileNameUniqueViolation` pattern already used in `AddFileController` for the same class of TOCTOU.
- `tests/NzbWebDAV.Tests/Api/CreateAccountControllerTests.cs` — 4 new tests: first-admin succeeds, second-admin rejected by the pre-check, a real race between two `DbContext`s against the actual SQLite unique index (only one admin survives), and confirmation that multiple `WebDav`-type accounts (an intentional, unrelated feature) are unaffected.

## Note on the migration snapshot diff

`DavDatabaseContextModelSnapshot.cs` shows a larger diff than the actual change - `dotnet ef migrations add` regenerated the whole file and reordered some pre-existing entity blocks in the process (confirmed by normalizing whitespace and diffing sorted lines against the pre-change version: the only real content differences are the new index block and a `ProductVersion` annotation bump from `10.0.4` to `10.0.10`, reflecting this environment's EF tooling version). No other entity content changed.

## Test plan

- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release` → 1420/1420 passing (1416 pre-existing + 4 new)
- [x] `dotnet build backend/NzbWebDAV.csproj -c Release` → 0 errors
- [x] New race test directly exercises the DB-level unique index with two separate `DbContext`s, not just the application-level pre-check

### DIFF
--- a/backend/Api/Controllers/CreateAccount/CreateAccountController.cs
+++ b/backend/Api/Controllers/CreateAccount/CreateAccountController.cs
@@ -1,4 +1,7 @@
-﻿using Microsoft.AspNetCore.Mvc;
+﻿using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Data.Sqlite;
+using Microsoft.EntityFrameworkCore;
 using NzbWebDAV.Database;
 using NzbWebDAV.Database.Models;
 using NzbWebDAV.Utils;
@@ -9,8 +12,23 @@ namespace NzbWebDAV.Api.Controllers.CreateAccount;
 [Route("api/create-account")]
 public class CreateAccountController(DavDatabaseClient dbClient) : BaseApiController
 {
-    private async Task<CreateAccountResponse> CreateAccount(CreateAccountRequest request)
+    internal async Task<CreateAccountResponse> CreateAccount(CreateAccountRequest request)
     {
+        // Onboarding's own isOnboarding()-then-createAccount() check in the frontend is not
+        // race-safe on its own: two concurrent onboarding submissions can both observe "no
+        // admin yet" before either insert commits. Re-check here, immediately before the
+        // write, and let the DB's own unique filtered index (IX_Accounts_SingleAdmin) be the
+        // authoritative backstop against the remaining race window between this check and
+        // SaveChangesAsync.
+        if (request.Type == Account.AccountType.Admin)
+        {
+            var adminExists = await dbClient.Ctx.Accounts
+                .AnyAsync(a => a.Type == Account.AccountType.Admin, HttpContext.RequestAborted)
+                .ConfigureAwait(false);
+            if (adminExists)
+                throw new BadHttpRequestException("An admin account already exists.");
+        }
+
         var randomSalt = Guid.NewGuid().ToString("N");
         var account = new Account()
         {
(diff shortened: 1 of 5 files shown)


hoivikaj:
> [!NOTE]
> This review was produced by a code-review agent acting on behalf of the maintainer.

## PR review: #740 — fix(auth): close TOCTOU race allowing multiple admin accounts

**Verdict:** Request changes

### Summary
The PR adds a filtered unique index (`IX_Accounts_SingleAdmin`) so at most one `Admin` account row can exist, plus a controller pre-check (clean 400) and a `DbUpdateException` catch for the residual race window. The claimed vulnerability is real on `main`, the fix is well-layered (DB index as the authoritative backstop, not just an app-level check), and the tests genuinely exercise the real SQLite index via two separate `DbContext`s. However, the commit is missing the required breaking-change marker for a migration, and the migration will boot-loop any instance that already contains duplicate admin rows — the exact state this PR exists to prevent.

### Claim validation
- "Only one admin, only during onboarding, enforced solely by frontend check-then-act" → **confirmed** — `/onboarding` and `/onboarding.data` are on the public allowlist in `frontend/app/auth/auth-middleware.server.ts`, the onboarding action does `isOnboarding()` then `createAccount()`, and pre-PR `CreateAccountController.CreateAccount` inserts unconditionally with no admin-existence check anywhere in the backend.
- "Two concurrent submissions could both insert an Admin row" → **confirmed** — `Accounts` PK is `(Type, Username)`, so two admins with different usernames coexist happily on pre-PR `main`. `IsOnboardingController` only checks existence, giving a cross-process TOCTOU.
- "Follows the existing `IsCategoryFileNameUniqueViolation` pattern" → **confirmed** — same shape as `AddFileController`.
- "Snapshot diff is only reordering + ProductVersion bump" → **confirmed** — the backend pins EF `10.0.10` in `backend/NzbWebDAV.csproj`, so the `10.0.4 → 10.0.10` snapshot bump is legitimate catch-up; the only content change in the Account block is the new index.
- "Multiple WebDav accounts unaffected" → **confirmed** — index filter is `"Type" = 1` and `Account.AccountType.Admin = 1`; only `CreateAccountController` ever inserts `Accounts` rows, so no other flow can trip the index.
- "1420/1420 tests passing" → **unverified locally** — and no CI checks have run on the branch yet (fork workflow approval likely needed).

### Must-fix
- **Missing breaking-change marker.** The commit and PR title are plain `fix(auth):`, but the PR adds a migration under `backend/Database/Migrations/`. Per AGENTS.md this must be `fix(auth)!:` (or carry a `BREAKING CHANGE:` footer) so release-please bumps as breaking and operators are told to back up `/config`.
- **Migration hard-fails on databases that already have 2+ admin rows.** `CREATE UNIQUE INDEX` in `20260731171110_Add-SingleAdmin-UniqueIndex.cs` will throw if duplicates exist — precisely the state produced by the bug being fixed (or by manual DB edits). Since migrations auto-apply on startup, an affected instance boot-loops with no recovery path. `Up()` should first resolve duplicates (e.g. keep the earliest-`rowid` admin and delete the rest, with a logged warning) before creating the index — there is precedent for data-fixing migrations in this repo (`Fix-Empty-Categories`).
- **`IsSingleAdminUniqueViolation` misclassifies primary-key violations.** A duplicate-username insert violates the composite PK and SQLite reports `UNIQUE constraint failed: Accounts.Type, Accounts.Username`, which matches the `"UNIQUE"` + `"Accounts.Type"` fallback heuristic — so creating a second `WebDav` account with an existing username would return the misleading "An admin account already exists." Guard the catch with `request.Type == Account.AccountType.Admin` (mirroring the pre-check) or tighten the message match to exclude `Accounts.Username`.

### Nice-to-have
- A duplicate-username `WebDav` insert still surfaces as a raw 500 with a `Log.Error` stack (pre-existing, and adjacent to this code) — a friendly 400 for PK violations would follow the AGENTS.md stack-dump guidance, but it's out of scope if you want to keep the diff minimal.

### Notes
- Tests: adequate — 4 focused tests including a real two-context race against the actual SQLite index, using established test-suite patterns (`ConfigPathCollection`, migration-backed SQLite, `InternalsVisibleTo` already in place).
- Breaking / migration: **yes, migration included — commit marker NOT OK** (see Must-fix).
- Residual risk: no CI has run on this branch yet — approve the fork workflow run before merging. `CHANGELOG.md` untouched, no secrets, no proxy-allowlist changes needed.

hoivikaj:
Addressed all review items in 60d5f0ae:

- repairs pre-existing duplicate admins before creating the unique index, retaining the earliest account
- distinguishes the single-admin index from composite primary-key violations
- returns a friendly 400 for duplicate usernames
- adds migration-repair and duplicate-username regression tests
- marks the migration as breaking in both the commit and PR title

Validation: `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release` (1413 passed, 14 skipped).

PR #746, opened 2026-08-02
fix(usenet): body-mode playback verification no longer leaks provider connections

loambit:
Found this bug while looking into something else. With `play.verify-mode=body`, PlaybackFastVerifier fetches the article and reads only the response code, so the decoded body is never released and the connection behind the probe stays out for the life of the process. The verdict is now computed before the body is released, and the release is guarded so a failure returning it cannot turn a verified article into a failed one.


hoivikaj:
@loambit thanks!

PR #775, opened 2026-08-03
fix(ui): selection ticks, swallowed clicks, and a hidden menu in the file browser

loambit:
Fixes these UI bugs on the Files section:
- The row you just clicked showed no checkmark, even though it was selected and counted.
- Clicking outside an open item menu closed it but swallowed that click, so the thing you clicked needed a second click.
- Selecting an option didn't close the menu, most visibly leaving it open behind the Remove confirm modal.
- An open menu was invisible on the last row of a listing, so on a single-file folder it couldn't be seen at all.


hoivikaj:
@loambit thank you sir!

PR #818, opened 2026-08-05
feat(ui): native URL_BASE support for sub-path hosting

STiXzoOR:
## What

Adds first-class support for hosting the app under a reverse-proxy sub-path — e.g. `https://example.com/nzbdav/` — without any `sub_filter` response rewriting in the proxy. Deployments that share one domain across many services (Swizzin/organizr-style setups, corporate proxies) currently need ~15 fragile nginx rewrite rules to keep the SSR app, SAB API, WebSocket, and hashed asset paths internally consistent; this replaces all of that with a single `URL_BASE` value.

## Design

`URL_BASE` has two halves that must match (documented in `docs/configuration/url-base.md`):

- **Build time** (`--build-arg URL_BASE=/nzbdav`): React Router basename, Vite `base`, and a `__URL_BASE__` constant baked into the client bundle. React Router's basename is build-time only — the framework exposes no runtime override — so this half cannot be avoided. The Dockerfile defaults the runtime env var to the baked value, so build-arg-only is enough.
- **Runtime** (`URL_BASE` env): `server.ts` mounts static assets and the app module on a sub-router under the prefix, the WebSocket endpoint moves to `<URL_BASE>/ws`, and login redirect Locations get the prefix restored.

Key property: inside the mounted router `req.path` is stripped of the prefix, so the existing backend-proxy path checks, credential rate-limiter paths, and auth public paths all work **unchanged** — while the React Router request handler reads `req.originalUrl` and still sees the full basename-prefixed path. The backend stays completely prefix-unaware (rclone and `NZBDAV_CONFIG__…` consumers are unaffected).

- `GET /healthz` stays at the bare root (stable container healthchecks) and is additionally served under the prefix.
- Bare `/` redirects to `<URL_BASE>/`.
- Client code uses a new `~/utils/url-base` helper (`withUrlBase`) instead of hand-rolled prefix arithmetic; `<Link>`/`navigate()` need no changes since the basename covers them.
- **Unset `URL_BASE` produces output identical to before this change** — the published root-hosted images are unaffected.

## Testing

- New unit tests for the normalizer/joiner (`url-base.test.ts`); full suite passes (507 tests), `typecheck` and `lint` clean.
- Built with `--build-arg URL_BASE=/nzbdav` and functionally verified end-to-end on a migrated production-scale database: bare-root redirect, unauthenticated redirect to `<URL_BASE>/login`, login POST → 302 `<URL_BASE>/`, authenticated SSR pages with all asset/manifest URLs prefixed, SAB API under `<URL_BASE>/api`, WebSocket 101 on `<URL_BASE>/ws`, WebDAV PROPFIND under the prefix, and root+prefixed `/healthz`.
- Also verified a default build (no `URL_BASE`) behaves identically to current `main`.

This mirrors a patch we've run in production since May on the nzbdav-dev lineage (STiXzoOR/nzbdav) and have now ported/re-validated on current `main`. Happy to adjust naming (`URL_BASE` vs `general.base-url` interplay), split commits, or add CI matrix coverage for a sub-path build if you'd take it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,24 @@ jobs:
         working-directory: frontend
         run: npm run build && npm run build:server
 
+      # Sub-path build: the only automated coverage NZBDAV_URL_BASE can get.
+      # Asserts the baked bundle carries the prefix in its asset URLs and that
+      # the root build above was produced without one.
+      - name: Build with NZBDAV_URL_BASE
+        working-directory: frontend
+        env:
+          NZBDAV_URL_BASE: /nzbdav
+        run: |
+          mv build build-root
+          npm run build
+          if ! grep -rq '/nzbdav/assets/' build/server; then
+            echo "sub-path build did not bake /nzbdav into the bundle"; exit 1
+          fi
+          if grep -rq '/nzbdav/assets/' build-root/server; then
+            echo "root build unexpectedly contains the /nzbdav prefix"; exit 1
+          fi
+          rm -rf build && mv build-root build
+
       # Node ESM requires explicit .js extensions in emitted dist-node imports.
       # Import every helper under dist-node/server/ (not server.js, which listens).
       - name: Smoke-import dist-node server modules
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,6 +7,13 @@ FROM --platform=$BUILDPLATFORM node:24-alpine AS frontend-build
 
 WORKDIR /frontend
 
+# URL_BASE bakes the reverse-proxy sub-path (e.g. "/nzbdav") into the React
+# Router basename, Vite asset paths, and the client bundle. Leave unset for
+# root hosting (default, identical output to before this arg existed). Must
+# match the URL_BASE env var at container runtime — see docs/configuration/url-base.md.
+ARG URL_BASE=""
+ENV URL_BASE=${URL_BASE}
+
 COPY ./frontend/package.json ./frontend/package-lock.json ./
 RUN npm ci
 COPY ./frontend ./
@@ -86,5 +93,10 @@ ARG NZBDAV_COMMIT_SHA
 ENV NZBDAV_COMMIT_SHA=${NZBDAV_COMMIT_SHA}
 ENV NODE_ENV=production
 ENV LOG_LEVEL=warning
+# Default the runtime half of URL_BASE to the baked build-time value so a
+# `--build-arg URL_BASE=/x` image works without repeating the env var at run
+# time. Can still be overridden (must then match the build value).
+ARG URL_BASE=""
+ENV URL_BASE=${URL_BASE}
 
 CMD ["/entrypoint.sh"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,12 +7,14 @@ FROM --platform=$BUILDPLATFORM node:24-alpine AS frontend-build
 
 WORKDIR /frontend
 
-# URL_BASE bakes the reverse-proxy sub-path (e.g. "/nzbdav") into the React
-# Router basename, Vite asset paths, and the client bundle. Leave unset for
-# root hosting (default, identical output to before this arg existed). Must
-# match the URL_BASE env var at container runtime — see docs/configuration/url-base.md.
-ARG URL_BASE=""
-ENV URL_BASE=${URL_BASE}
+# NZBDAV_URL_BASE bakes the reverse-proxy sub-path (e.g. "/nzbdav") into the
+# React Router basename, Vite asset paths, and the client bundle. Leave unset
+# for root hosting (default, identical output to before this arg existed).
+# Namespaced to avoid picking up a stray URL_BASE from shared build
+# environments; the un-prefixed name is still honored as a fallback by the
+# build config. See docs/configuration/url-base.md.
+ARG NZBDAV_URL_BASE=""
+ENV NZBDAV_URL_BASE=${NZBDAV_URL_BASE}
 
 COPY ./frontend/package.json ./frontend/package-lock.json ./
 RUN npm ci
@@ -93,10 +95,10 @@ ARG NZBDAV_COMMIT_SHA
 ENV NZBDAV_COMMIT_SHA=${NZBDAV_COMMIT_SHA}
 ENV NODE_ENV=production
 ENV LOG_LEVEL=warning
-# Default the runtime half of URL_BASE to the baked build-time value so a
-# `--build-arg URL_BASE=/x` image works without repeating the env var at run
-# time. Can still be overridden (must then match the build value).
-ARG URL_BASE=""
-ENV URL_BASE=${URL_BASE}
+# Default the runtime half of NZBDAV_URL_BASE to the baked build-time value so
+# a `--build-arg NZBDAV_URL_BASE=/x` image works without repeating the env var
+# at run time. Overriding it with a different value fails fast at startup.
+ARG NZBDAV_URL_BASE=""
+ENV NZBDAV_URL_BASE=${NZBDAV_URL_BASE}
 
 CMD ["/entrypoint.sh"]
--- a/docs/configuration/url-base.md
+++ b/docs/configuration/url-base.md
@@ -1,44 +1,62 @@
-# URL base (sub-path hosting)
+# URL base (sub-path hosting) [since 0.11.0](https://github.com/nzbdav/nzbdav/releases/tag/v0.11.0){ .nzbdav-since }
 
 Host the app under a reverse-proxy sub-path — e.g. `https://example.com/nzbdav/` —
 without any `sub_filter` response rewriting in the proxy.
 
-The setting is the `URL_BASE` value and it has **two halves that must match**:
+The setting is the `NZBDAV_URL_BASE` value (bare `URL_BASE` is accepted as a
+fallback) and it has **two halves that must match**:
 
 | Half | When | What it controls |
 | --- | --- | --- |
-| `--build-arg URL_BASE=/nzbdav` | Docker image build | React Router basename, Vite asset paths, and the `__URL_BASE__` constant baked into the client bundle |
-| `-e URL_BASE=/nzbdav` | Container runtime | The Express mount prefix, WebSocket endpoint path, and login-redirect Locations |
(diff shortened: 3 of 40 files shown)


hoivikaj:
> [!NOTE]
> I am a Code Review agent operating under the context of the repository maintainer. Findings below were verified against the diff and against the actual dependency sources pinned in this repo (Express 5 / `router` v2, `compression`, `http-proxy-middleware` v4 / httpxy) rather than assumed from docs.

**Verdict: request changes.** The core design holds up under verification — the concerns are about hardening the env-var surface before this ships.

### Verified working as claimed

- Unset `URL_BASE` output is a true no-op: `withUrlBase()` is an identity, basename and Vite base stay `/`, the WebSocket path stays `/ws`, and the router mounts at root in the same relative order as before. Published root-hosted images are unaffected.
- The backend really does stay prefix-unaware: httpxy builds the outgoing path from the mount-stripped `req.url`, so the `/api`/WebDAV proxy checks, rate-limiter paths, and auth public paths work unchanged, exactly as described.
- The compression skip list keeps working under a prefix (checked because it *looks* broken at first glance): `compression` evaluates its filter at `onHeaders` time, after Express has stripped the mount prefix from `req.url`, and the proxy handler never calls `next()` to restore it. Not an issue.

### Must fix

1. **Namespace the variable — `NZBDAV_URL_BASE`** (optionally accepting bare `URL_BASE` as a fallback). `URL_BASE` is the exact name SABnzbd/Arr users already know, and both `vite.config.ts` and `react-router.config.ts` read it from the ambient environment at build time. Deployments that build from source inside a shared container environment — DUMB is the flagship example; it merges per-service `env` with the container environment during builds — would silently bake a basename from a stray value intended for another service, breaking DUMB's root-style embedded-UI proxying with no error anywhere. Everything else here is already namespaced (`NZBDAV_CONFIG__*`, `NZBDAV_COMMIT_SHA`).

2. **Guard the build/runtime split.** `server.ts` reads only the runtime env var; nothing compares it against the baked `__URL_BASE__`. Setting `URL_BASE` on a root-built image (or after a source install was built) activates the runtime half only: the server mounts under the prefix while the bundle and basename sit at root — a fully broken app with zero diagnostic. The server should obtain the baked value (import from the SSR bundle, or emit it during build) and refuse to start — or at minimum log one clear operator-facing line — on mismatch.

3. **Validate the value before mounting.** `app.use(URL_BASE, router)` hands the raw string to path-to-regexp v8. Verified behavior: `(` or `{` crashes at boot with a raw stack (`Unexpected ( at index 4: …`), while `:` and `*` mount *successfully* but silently turn the prefix into a pattern — `URL_BASE=/nzb:dav` serves the app under `/nzbanything`. `normalizeUrlBase` should reject anything outside a safe path charset with a clear startup error.

### Should fix

4. `securityHeadersMiddleware` sits on the root app above the mount and sets headers synchronously, so under a prefix its `shouldProxyToBackend(req.path)` exemption never matches (`/nzbdav/view/…`) and proxied WebDAV/API **GET** responses gain the `nosniff`/`X-Frame-Options`/`Referrer-Policy` headers the code deliberately withholds (PROPFIND/OPTIONS stay exempt via the method check, and httpxy does not clear pre-set headers). Low practical impact, but it silently breaks that file's stated contract — move it inside the mounted router or make the check prefix-aware.

5. `public/site.webmanifest` has root-absolute icon `src` values and no `start_url`/`scope`. Under a prefix the manifest itself is served at `<URL_BASE>/site.webmanifest`, but its icons still point at `/android-chrome-*` and 404 through the proxy — contradicting the "all asset/manifest URLs prefixed" claim.

6. Only `GET /` is redirected; a `PROPFIND /` 404s. Anyone with rclone pointed at the **frontend** port root — the common wiring, and what DUMB configures automatically — breaks the moment `URL_BASE` is set. The doc only covers "rclone pointed at the backend port"; please state explicitly that frontend-port WebDAV consumers must include the prefix in their remote URL, and that managed environments which assume root hosting (e.g. DUMB) should not enable this setting.

### Nits

- `normalizeUrlBase` now exists in five hand-synced copies (`app/utils/url-base.ts`, `react-router.config.ts`, `vite.config.ts`, `server.ts`, `server/oidc-routes.ts`). At minimum add a parity test; some copies could import a shared non-Vite module.
- The new tests only exercise the empty prefix (the test file's own comment acknowledges this). Your offer of a CI matrix build with `--build-arg URL_BASE=/nzbdav` is worth taking — it is the only automated coverage this feature can realistically get.
- Docs: the new page heading needs a `since` version pill per repo convention.
- Title/commit should carry a scope (`feat(ui): …`).

Thanks for the thorough writeup and testing notes — the mount-the-router approach with `req.path` stripping is clean, and the claims that could be verified mechanically all checked out.

STiXzoOR:
Thanks for the thorough, verified review — all findings confirmed on our side too. Everything is addressed in `cf0f2bdb` (and the original commit is reworded to `feat(ui):` per the nit; history rewritten accordingly).

**Must fix**

1. **Namespaced** → `NZBDAV_URL_BASE`, with bare `URL_BASE` still honored as a documented fallback (`urlBaseFromEnv()` reads `NZBDAV_URL_BASE ?? URL_BASE`). Dockerfile arg renamed to match. Good catch on the DUMB shared-env build hazard — the fallback keeps single-service source builds convenient while the namespaced name wins whenever both are present.
2. **Build/runtime guard** → the SSR bundle now exports `bakedUrlBase` (the `__URL_BASE__` value it was compiled with) and `server.ts` calls `assertUrlBaseMatchesBuild()` before mounting — in both prod and dev paths. On mismatch it logs one operator-facing line naming both values and the two ways to fix it, then exits 1. Verified live: a root-built image started with `NZBDAV_URL_BASE=/nzbdav` now refuses to start with `NZBDAV_URL_BASE mismatch: this build was compiled with "" but the runtime environment says "/nzbdav" …`.
3. **Validation** → `normalizeUrlBase` now rejects anything outside `[A-Za-z0-9._~\-/]` with a clear error (tests cover `:`, `*`, `(`, `{`, space, `%`, `?`). Reproduced your finding first: `:`/`*` did silently mount as a pattern.

**Should fix**

4. `securityHeadersMiddleware` moved inside the mounted router, so its `shouldProxyToBackend(req.path)` exemption sees stripped paths again. Verified under `/nzbdav`: UI responses carry `nosniff`/`X-Frame-Options`, proxied `/api` GETs no longer do. (Side effect worth noting: `/healthz` no longer gets the headers — it returns before them anyway in spirit, but say the word if you'd rather keep them there.)
5. `site.webmanifest` → relative icon `src` values plus `start_url`/`scope` of `"./"`, all resolving against the manifest's own URL — prefix-agnostic with no build-time rewriting.
6. Docs now state explicitly that only `GET /` redirects, that WebDAV consumers pointed at the **frontend** port must include the prefix in their remote URL, and carry a warning that managed root-hosting environments (DUMB et al.) should leave the setting unset.

**Nits**

- Copies collapsed 5 → 2: `vite.config.ts`, `react-router.config.ts`, `oidc-routes.ts`, and `auth-middleware.server.ts` now import the shared helper; `server.ts` (tsc → dist-node, outside the app graph) keeps a mirror with a parity test (`url-base.test.ts`) asserting the guard regex and normalization steps appear verbatim in both files, plus that `server.ts` reads the namespaced env var.
- CI now builds with `NZBDAV_URL_BASE=/nzbdav` and asserts the prefix is baked into the sub-path bundle and absent from the root bundle.
- Docs heading has the since-pill (guessed `0.11.0` — happy to adjust to whatever release this lands in).

Full test suite: 517 passing (was 507), typecheck and lint clean. Re-ran the end-to-end battery on a production-scale database with the rebuilt image: root+prefixed `/healthz`, bare-root redirect, login POST → 302 under the prefix, assets/manifest prefixed, SAB API, WebSocket 101.

hoivikaj:
> [!NOTE]
> Code Review agent again, operating under the context of the repository maintainer.

**Verdict: Approve** (with one residual note on the since-pill version).

### Merge conflict

`main` had moved ahead by #819 / #820 and conflicted in `frontend/app/routes/settings/indexers/indexers.tsx`. I merged `main` into this branch and pushed `681d8452` (maintainer-can-modify). Resolution keeps `withUrlBase` for the sub-path fetches and retains #819's removal of the masked-secret Test gate (backend now resolves saved keys). PR is `MERGEABLE` again.

### Follow-up claim check (`cf0f2bdb`)

| Prior must-fix | Status |
| --- | --- |
| Namespace → `NZBDAV_URL_BASE` (+ `URL_BASE` fallback) | **done** — `urlBaseFromEnv`, Dockerfile arg, shared-env hazard covered |
| Build/runtime mismatch guard | **done** — `bakedUrlBase` export + `assertUrlBaseMatchesBuild()` in both prod and dev; exits 1 with an actionable line |
| Validate before mount | **done** — charset guard + tests for `:`, `*`, `(`, `{`, space, `%`, `?` |
| Move `securityHeadersMiddleware` inside the mounted router | **done** |
| Prefix-agnostic `site.webmanifest` | **done** — relative icons + `start_url`/`scope` of `"./"` |
| Frontend-port WebDAV / managed-env docs | **done** — non-GET root note + DUMB warning |

Nits from last round also landed: 5→2 normalizer copies with a parity test, CI sub-path build assertion, `feat(ui):` scope.

### Residual note

- Docs since-pill is `0.11.0` — fine if that is the intended landing release; adjust if release-please is still on a different line when this merges.
- `/healthz` no longer getting security headers (called out in your reply) is fine — probes do not need them.

Thanks for the thorough turnaround. Ready from the review side once CI is green.

STiXzoOR:
CI note: the run on `681d8452` came back with a single failure — `macOS yEnc / library tests`, specifically `UsenetSharp DecodedBodiesAsync_CancellationDrainsBatchAndReusesConnection` (1 of 227; every other job green, including the new sub-path build assertion in the frontend job).

This PR's diff touches only `frontend/`, `docs/`, `zensical.toml`, the `Dockerfile`, and the frontend section of `ci.yml` — nothing under `libs/UsenetSharp` or the backend — and the test is a cancellation-race with tight timing on a shared macOS runner, so this looks like an environment flake rather than anything introduced here. I don't have permission to re-run jobs on this repo; could you kick a re-run of the failed job when you get a chance?

PR #982, opened 2026-08-13
fix(webdav): /ready no longer returns 4XX instead of the readiness status

loambit:
Adds /ready to the WebDAV path filter and the frontend's backend proxy prefixes so the endpoint reaches StreamingReadinessCheck instead of returning 401 on the backend port and 404 on the frontend one.


hoivikaj:
@loambit Thanks for this — the diagnosis and the denylist/proxy wiring are right. We're pushing a small adjustment onto this branch: an HTTP pipeline test that `GET /ready` returns 200 without WebDAV auth, matching the existing `/health` coverage. The filter unit tests wouldn't have caught middleware still challenging the route.

Issue #1043, opened 2026-08-19
Close remaining in-flight budget gaps: unbudgeted seek-head buffer and retry-under-saturation stall


hoivikaj:
Identified during the #783 review/implementation (PR pending). Two related gaps in in-flight article budget coverage/ordering:

**1. Unbudgeted seek-head buffer.** `NzbFileStream.TryGetSeekStreamFast` (`backend/Streams/NzbFileStream.cs`, head drain ~lines 395-413) drains one segment tail into a `PooledBufferStream` per seek **without an `ArticleByteLease`**. Since the pipe-accounting fix, the seek head's pipe bytes count against the budget, but the pooled head buffer itself remains uncounted.

Invariant for any fix (from the deadlock-safety contract): the lease must be acquired **before** `DecodedBodyAsync` using an estimate — never after the pipe exists. Leasing with an open pipe is the one shape the contract forbids (a `LeaseAsync` waiter must never hold an open body pipe).

**2. Pre-existing retry-under-saturation window** (empirically confirmed during #783 test development; exists on main today): with a tiny budget, a multi-segment prefetch window, and a mid-file retry, the retry's re-lease queues behind the producer's earlier-enqueued waiter, while later drained-but-unread buffers hold leases the ordered consumer cannot release until it passes the retrying segment — a stall.

Repro shape: budget ≈ 1.25 segments, `articleBufferSize: 4`, one transient failure mid-file (the pre-split combined test from the #783 branch history hung for 30s on exactly this).

Neither is a regression — both predate the pipe-accounting change, which only shifts saturation onset marginally earlier (≤ ~1.09 MiB × connections).

PR #1052, opened 2026-08-20
fix(webdav): paused or trickling clients no longer wedge all streaming

### LINKED ISSUE #1041: In-flight article budget: leases held by a stalled/paused drain starve every other stream — saturation deadlocks instead of throttling
### Before submitting

### InfiniDysk version

ghcr.io/infinidysk/infinidysk:1.1.1 (code unchanged through v1.2.0-rc.1 / current main)

### Installation method

Official Docker image

### Affected area

WebDAV or streaming

### What happened?

`InFlightArticleBudget` (backend/Streams/InFlightArticleBudget.cs) caps decoded article bytes process-wide. A lease is released only when `ArticleByteLease.Dispose()` runs — i.e. when the **client** finishes draining that segment. There is **no reclaim or timeout path** for a consumer that stops draining.

A paused Plex client is enough: the player stops reading its WebDAV stream mid-segment but keeps the connection open, so its decoded articles stay leased indefinitely. Once one or two such drains pin `Leased` near `Cap`, every other stream on the server blocks inside `LeaseAsync`'s FIFO wait queue. Nothing errors, nothing times out, nothing logs beyond one throttled warning per 30s:

```text
[WRN] In-flight article memory budget saturated. Leased=1,073,704,489 Cap=1,073,741,824 Requested=4,193,911. Reason: backpressure
```

Observed effect on a live box (captured during the incident): an active 4K remux read froze at a single byte offset for 10+ minutes; every NNTP connection was healthy but idle (the pool goes quiet because no stream can lease memory to fetch into); `/healthz` and admin endpoints stayed green; only tearing the mount/app down recovered it. From the outside the server is "up" but globally wedged for streaming — backpressure that should degrade one stream instead deadlocks all of them.

Secondary hardening note found while reading the same file: `SignalWaiters()` is head-only and edge-triggered — it calls `TrySetResult` on the head waiter's **current** TCS, while the waiter swaps in a fresh TCS (`Waiter.Reset()`) each time around its retry loop. A `Release` that fires in the window between a waiter's wake-up and its `Reset()` lands on the already-completed prior TCS and is a no-op — a lost wake-up. It only matters when the freed bytes still don't satisfy the head (so it re-waits) and no further releases arrive — which is exactly the regime the primary bug creates.

### Steps to reproduce

1. Set `usenet.in-flight-article-budget-mb` low (e.g. 64) to make it easy to pin.
2. Open stream A via WebDAV (large file), read a few MB, then **stop reading without closing the connection** (a paused media player, or `curl --limit-rate 1` piped to a paused consumer).
3. Open additional streams until the saturation warning fires and `Leased` sits at/near `Cap`.
4. Attempt any further stream: it blocks in `LeaseAsync` indefinitely. No timeout, no error, no per-stream failure — for hours, as long as A stays paused.
5. Resume or kill stream A: everything instantly recovers — confirming the only wedge was the un-reclaimable leases.

### Expected behavior

Backpressure should cost the stalled stream, not the whole server:

- **Lease reclaim:** if a drain has made no progress for N seconds while others are blocked on admission (`_waiters` non-empty), abort that response and release its leases. The stalled client gets a broken stream it can retry; every other stream keeps flowing. (Precedent: the per-segment fetch path already fail-fasts; it is only the drain side that can hold memory forever.)
- **Level-triggered wake-ups:** record releases as a pending permit/generation counter checked under `_gate` before each `await`, so a signal can never be lost to a stale TCS.

Related: #800 sized the default cap; this issue is about what happens when the cap is reached — today the answer is "deadlock until the pauser goes away," which makes any cap value a wedge threshold rather than a safety limit.

```text
[WRN] In-flight article memory budget saturated. Leased=1,073,704,489 Cap=1,073,741,824 Requested=4,193,911. Reason: backpressure
# ...then silence: no per-stream errors, no repair activity, NNTP sockets healthy-idle,
# one reader pinned at a fixed byte offset 10+ minutes until external recovery.
```

### Environment

OS: Linux x86_64 (Docker)
Browser or WebDAV client: rclone VFS mount (consumed by Plex); also reproducible with curl

### Additional context

_No response_

hoivikaj:
## Summary

A paused or few-KB/s client could pin its prefetched Article RAM leases indefinitely: each 64 KB write still finished inside the per-write deadline, so the existing watchdog never fired. The process-wide in-flight article budget then saturated and every other stream blocked in `LeaseAsync` until restart (#1041). `/healthz` stayed green because the stalled session was still an active read.

This change:

- Cancels a stream when write-side throughput stays below 64 KB per timeout window **and** another stream is waiting on the article budget (idle slow clients are left alone).
- Applies the same per-write watchdog to `/view` (it previously had none).
- Stops newcomers barging past a queued FIFO waiter when capacity is freed.
- Reuses `usenet.streaming-write-timeout-seconds` (0 still disables both checks) and updates the setting help text.

The claimed lost-wake TCS race is not reachable: `TryLease` and `Waiter.Reset()` share the same lock as `SignalWaiters`. A generation counter is not added.

Fixes #1041

## Test plan

- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release` (3211 passed)
- [x] Trickle-reader reclaim: contended budget + <64 KB over the window cancels the linked token; uncontended does not
- [x] `/view`/shared helper stalled-client watchdog (never-completing write stream)
- [x] FIFO barging: queued head waiter gets released capacity before a newcomer
- [ ] Manual: low article budget + paused and trickling clients over WebDAV and `/view`; reclaim fires only under contention; warning is the throttled single-line event; streaming recovers without restart

Issue #1043 (unbudgeted seek-head buffer and retry-under-saturation) is a follow-up PR on top of this branch.


hoivikaj:
Fable 5 review follow-up (now on latest `main`, including shared streams):

- Record `local-cap-wait` latency on the in-lock lease success path (contended admissions were silently dropping that phase).
- Aggregate reclaim uses `CancelAsync` instead of synchronous `Cancel()`, so a throwing callback cannot replace `StreamingWriteTimeoutException` with `AggregateException`.
- Trickle reclaim logs `Reason: streaming-write-reclaim` with a distinct warning; per-write stalls keep `streaming-write-timeout`.

PR #1053, opened 2026-08-20
fix(streams): seeking and article retries no longer stall when Article RAM is full

### LINKED ISSUE #1043: Close remaining in-flight budget gaps: unbudgeted seek-head buffer and retry-under-saturation stall

hoivikaj:
## Summary

Follow-up to #1052 / #1041. Two separate budget gaps from #1043:

1. **Unbudgeted seek-head buffer.** Fast seek drained the target-segment tail into a `PooledBufferStream` without an `ArticleByteLease`, so that RAM sat outside the process cap.
2. **Retry-under-saturation stall.** On a transient failure the producer disposed the segment lease and re-leased, which queued the retry behind already-enqueued waiters while drained-but-unread buffers still held capacity. With a tiny cap (≈1.25 segments) and a prefetch window of 4, `CopyToAsync` hung.

This change:

- Acquires an `ArticleByteLease` for the estimated tail **before** `DecodedBodyAsync` (a waiter must never hold an open body pipe), adjusts to the actual tail, and transfers ownership through `BudgetedStream` so `CombinedStream.OnDispose` releases it once.
- Keeps the original segment lease across retry, rescue, and fallback attempts. The entitlement is held through backoff; it still releases on terminal failure or zero-fill.

Stacked on #1052 because both PRs touch `InFlightArticleBudget` waiters/leases. Merge #1052 first, then retarget this PR to `main` (or merge this into the 1041 branch).

Fixes #1043

## Test plan

- [x] `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release` (3211 passed, including this branch)
- [x] After a fast seek, `LeasedBytes` covers the pooled head and returns to 0 on dispose (including seek-then-dispose before the first read)
- [x] Tiny-budget retry repro (budget ≈ 1.25 segments, `articleBufferSize: 4`, one transient mid-file failure) completes without hanging
- [ ] Manual: low budget + scrub/seek and a forced mid-file retry; streams complete without wedging the host


hoivikaj:
Fable 5 review follow-up, merged onto #1052 / `main`:

- Downward seek-guess correction no longer leases `ArticleByteLease.Empty` (exact ranges when present, otherwise one average segment, then `Adjust`).
- `FastSeek_DisposeBeforeFirstRead_DoesNotLeakLease` now calls `TryGetSeekStreamFast` and disposes the spliced stream before any read.
- Known-corrupt `GetCorruptionRetryLimit` and Par2 `ReportCorruption` are preserved in the restructured retry loops.
- Added a pipelined tiny-budget retry test on the `DownloadBatchSegment` → `TryRescue` path (batch width 1 so the test isolates lease retention, not batch-wide admission).

PR #1118, opened 2026-08-22
feat(usenet): make provider circuit-breaker cooldowns configurable

loambit:
Make circuit breaker cooldown settings configurable, matching existing defaults. On systems with only 1 primary this prevents the backup block providers from being drained too fast.

## Summary by CodeRabbit

* **New Features**
  * Added configurable initial and maximum cooldown durations for provider circuit breakers.
  * Cooldowns now increase exponentially after repeated failures and reset after successful recovery.
  * Invalid settings are validated and safely clamped to supported ranges.

* **Documentation**
  * Documented cooldown configuration, backoff behavior, limits, persistence, and recovery probes.

* **Tests**
  * Added coverage for cooldown escalation, resets, defaults, parsing, and boundary handling.

### DIFF
--- a/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
+++ b/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
@@ -60,7 +60,12 @@ public class ProviderCircuitBreaker
         _providerName = providerName;
         _onTransition = onTransition;
         _coalesceFailureBursts = coalesceFailureBursts;
-        _initialCooldown = initialCooldown ?? DefaultInitialCooldown;
+        // A non-positive initial cooldown would trip straight into half-open and the
+        // doubling ladder would never grow, silently disabling the breaker.
+        var configuredInitial = initialCooldown ?? DefaultInitialCooldown;
+        _initialCooldown = configuredInitial > TimeSpan.Zero
+            ? configuredInitial
+            : DefaultInitialCooldown;
         // A ceiling under the initial cooldown would shorten the first trip instead of
         // bounding the ladder, so keep it at or above the value it caps.
         var ceiling = maxCooldown ?? DefaultMaxCooldown;
--- a/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
+++ b/backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
@@ -27,20 +27,22 @@ public class ProviderCircuitBreaker
     private const int MinFailuresToTrip = 3;
     private const double TripFailureRate = 0.5;
 
-    private static readonly TimeSpan InitialCooldown = TimeSpan.FromSeconds(60);
-    private static readonly TimeSpan MaxCooldown = TimeSpan.FromMinutes(5);
+    private static readonly TimeSpan DefaultInitialCooldown = TimeSpan.FromSeconds(60);
+    private static readonly TimeSpan DefaultMaxCooldown = TimeSpan.FromMinutes(5);
     private static readonly TimeSpan FailureBurstCoalesceWindow = TimeSpan.FromSeconds(2);
     private static readonly TimeSpan DefaultProbeAbandonTimeout = TimeSpan.FromSeconds(60);
 
     private readonly string _providerName;
     private readonly Action<ProviderCircuitTransition>? _onTransition;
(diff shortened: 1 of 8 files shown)


backend/Config/ConfigManager.cs
```diff
@@ -1036,6 +1038,34 @@ public int GetWarmConnectionsFloor(int maxConnections)
...
+    /// <summary>
+    /// Cooldown applied the first time a provider's circuit trips. Each consecutive trip
+    /// doubles it up to the ceiling, and a successful body fetch resets it back to this
+    /// value. Defaults to 60s and is clamped to 5 through 300. The connection pools read
+    /// it when they are built, so a change applies on the next restart or provider save.
+    /// </summary>
+    public TimeSpan GetCircuitBreakerInitialCooldown()
+    {
+        var configured = StringUtil.EmptyToNull(GetConfigValue(ConfigKeys.UsenetCircuitBreakerInitialCooldownSeconds));
+        if (configured is null || !int.TryParse(configured, out var seconds))
```
hoivikaj: [P3] Parse the same numeric domain that validation accepts. These two keys are validated with `RequireLong`, but both getters use `int.TryParse`. For example, `2147483648` passes both the Settings API and `NZBDAV_CONFIG__...` startup validation, then resolves to the 60/300-second default instead of the documented 300/3600-second clamp. Please parse `long` in both getters (clamping before `TimeSpan` conversion), or restrict validation consistently, and add an above-Int32 boundary test.

docs/configuration/usenet.md
```diff
@@ -68,3 +68,25 @@ re-probing the same provider for the same article until the TTL expires. Transie
 | Miss-cache max entries | `usenet.article-miss-cache-max-entries` | `10000` | Cap before oldest entries are evicted (clamped 100–1000000) |
 
 The cache clears automatically when Usenet providers are reconfigured.
+
+## Provider circuit-breaker cooldown
```
hoivikaj: [P2] Add the required release marker for this new user-visible setting. Older images reject these new `NZBDAV_CONFIG__...` names as unknown during startup, so readers need the minimum supported version. Per the repository docs policy, please mark this heading with the anticipated `[since 1.3.0](https://github.com/infinidysk/infinidysk/releases/tag/v1.3.0){ .nzbdav-since }` pill.

hoivikaj (changes requested):
I am a code review agent operating under maintainer context.

The circuit-breaker wiring and default compatibility look sound, and the completed backend, documentation, and CodeQL checks pass. I’m requesting changes for the configuration parsing inconsistency and required documentation version marker noted inline. Backend mutation was still pending when this review was submitted.

backend/Config/ConfigManager.cs
hoivikaj: Fixed in c096749a2 — both getters now parse with `long.TryParse` and clamp before the `TimeSpan` conversion, and the boundary tests cover `2147483648` / `-2147483649`.

docs/configuration/usenet.md
hoivikaj: Fixed in c096749a2 — the heading now carries the `[since 1.3.0]{ .nzbdav-since }` pill.

backend/Clients/Usenet/Connections/ProviderCircuitBreaker.cs
```diff
@@ -51,11 +53,19 @@ public class ProviderCircuitBreaker
...
     {
         _providerName = providerName;
         _onTransition = onTransition;
         _coalesceFailureBursts = coalesceFailureBursts;
+        _initialCooldown = initialCooldown ?? DefaultInitialCooldown;
+        // A ceiling under the initial cooldown would shorten the first trip instead of
+        // bounding the ladder, so keep it at or above the value it caps.
+        var ceiling = maxCooldown ?? DefaultMaxCooldown;
+        _maxCooldown = ceiling < _initialCooldown ? _initialCooldown : ceiling;
+        _currentCooldown = _initialCooldown;
```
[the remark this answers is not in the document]
hoivikaj: Fixed in ec1fcb6f — a non-positive `initialCooldown` now falls back to the 60s default so the ladder can never start at zero, with theory cases covering 0 and -30s.

PR #1174, opened 2026-08-25
chore(dumb): expose a database contract file for SQLite-to-PostgreSQL migrators

### LINKED ISSUE #1150: Expose a machine-readable database contract for SQLite-to-PostgreSQL migrators
### Before submitting

@I-am-PUID-0

- [x] I have searched existing issues and did not find a similar request.
- [x] This request relates to use with legally obtained content.

### Problem or use case

DUMB (Debrid Unlimited Media Bridge) offers SQLite-to-PostgreSQL migration for InfiniDysk installs and keeps that migrator locked to specific schema versions.

The newest applied migration timestamp already helps, but it is not enough for a safe migration. DUMB checks the database’s **exact applied migration history** and the **resulting schema**, not just the newest migration file, because the installed source and the live database can differ (pending migrations, restored backups, forks, skipped/conflicting history).

Without a runtime contract from InfiniDysk, DUMB has to reverse-engineer schema identity from `__EFMigrationsHistory`, backup manifests, or app version — all of which can disagree.

### Proposed solution

Expose a machine-readable **database contract** from the InfiniDysk runtime, for example:

```json
{
  "contract": "infinidysk-db-v1.2.3",
  "provider": "sqlite",
  "terminalMigration": "20260822140000_Add-Profile-Stream-State-Index",
  "migrationCount": 50,
  "migrationHistoryHash": "...",
  "transientObjects": ["TMP_LINKED_FILES", "TMP_LINKED_FILES_UNIQUE"]
}
```

Requested fields:

- `contract` — stable identifier DUMB (and other migrators) can pin against
- `provider` — `sqlite` or `postgres`
- `terminalMigration` — newest applied EF migration id
- `migrationCount` — number of applied migrations
- `migrationHistoryHash` — fingerprint of the full applied history (not only the tip)
- `transientObjects` — objects that may exist at runtime but are not part of the stable schema (e.g. `TMP_LINKED_FILES` during unlinked-file cleanup)

Delivery can be an authenticated admin API, a file under `/config`, or another runtime-owned surface — whatever is stable for DUMB to read without scraping source.

### Alternatives considered

- **Newest migration id only** (backup `lastMainMigration`, support-pack last migration): insufficient when history and tip disagree.
- **App version** (`NZBDAV_VERSION` / `version.txt`): the binary and the database can be on different schemas.
- **Direct `__EFMigrationsHistory` reads:** possible for DUMB, but duplicates InfiniDysk’s source of truth and misses provider, transient objects, and a versioned contract id.
- **[#1012](https://github.com/infinidysk/infinidysk/issues/1012) in-app SQLite→Postgres migration:** complementary; DUMB still needs a pin-able contract for its own migrator.

### Relevant area

Other (database schema / DUMB integration)

### Additional context

Requested by the DUMB maintainer. InfiniDysk is a fully supported DUMB core module; PostgreSQL is currently documented as new-installs-only with no automatic SQLite→Postgres path ([docs/operations/postgresql.md](docs/operations/postgresql.md)).

`terminalMigration` in the example is illustrative of the shape they want (`20260822140000_Add-Profile-Stream-State-Index`); the live tip will move as new migrations land.

---

hoivikaj:
## Summary

- Writes a machine-readable database contract to `/config/db-contract.json` after every migration event (startup migrations, standalone `--db-migration`, and lazy creation/recreation of the usenet-migration ledger), so DUMB's SQLite-to-PostgreSQL migrator can pin against a stable runtime-owned contract instead of reverse-engineering `__EFMigrationsHistory`, backup manifests, or the app version.
- The contract carries a stable id (`infinidysk-db-v1`), provider, terminal migration, applied count, a `sha256` fingerprint of the full applied history, and known transient objects (`TMP_LINKED_FILES`, `TMP_LINKED_FILES_UNIQUE`) — top-level for the main database plus a `databases` map covering main, metrics, and usenet-migration.
- The file is world-readable (`0644`) and replaced atomically via temp-file + rename, so readers never see a partial document and a differently-owned leftover from a previous install can never block an upgrade or reinstall. Writes are best-effort and never block startup or the migration wizard.

Closes #1150

## Test plan

- [x] New `DatabaseContractWriterTests` (6 tests): contract shape against real migrated SQLite databases, hash determinism, 0644 permissions, overwrite of a read-only leftover, stale temp cleanup, lazily-created ledger handling, never-throws guarantee
- [x] Full backend suite: `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release` — 3713 passed, 0 failed
- [x] Docs build: `zensical build --clean --strict` — clean
- [x] Manual smoke: `NzbWebDAV --db-migration` against a temp CONFIG_PATH produced the expected `db-contract.json` with mode 0644
- [ ] CI green

## Summary by CodeRabbit

* **New Features**
  * Added automatic generation of a machine-readable database contract after database migrations.
  * Contract includes migration history, database metadata, application version, timestamps, and optional Usenet migration details.
  * Contract updates are written safely and do not block startup when recoverable errors occur.

* **Documentation**
  * Added operational documentation for `/config/db-contract.json`, including its format, permissions, contents, and update behavior.

### DIFF
--- a/backend/Database/DatabaseContractWriter.cs
+++ b/backend/Database/DatabaseContractWriter.cs
@@ -1,0 +1,179 @@
+using System.Security.Cryptography;
+using System.Text;
+using System.Text.Json;
+using Microsoft.EntityFrameworkCore;
+using NzbWebDAV.Config;
+using Serilog;
+
+namespace NzbWebDAV.Database;
+
+/// <summary>
+/// Writes a machine-readable database contract (<c>db-contract.json</c>) to CONFIG_PATH
+/// after migration events, so external migrators (e.g. DUMB's SQLite-to-PostgreSQL
+/// migrator) can pin against the exact applied migration history instead of
+/// reverse-engineering schema identity from __EFMigrationsHistory, backup manifests,
+/// or the app version. The top-level fields describe the main database; the
+/// <c>databases</c> map carries every database the runtime owns.
+/// </summary>
+internal static class DatabaseContractWriter
+{
+    internal const string ContractVersion = "infinidysk-db-v1";
+    internal const string ContractFileName = "db-contract.json";
+
+    // Runtime tables that are not part of the stable schema. RemoveUnlinkedFilesTask
+    // recreates TMP_LINKED_FILES on every orphan-cleanup pass, and TMP_LINKED_FILES_UNIQUE
+    // can be stranded when a run dies between CREATE and RENAME.
+    private static readonly string[] MainTransientObjects = ["TMP_LINKED_FILES", "TMP_LINKED_FILES_UNIQUE"];
+
+    // Test seams (same pattern as UsenetMigrationStore).
+    internal static Func<DavDatabaseContext> MainContextFactory { get; set; } =
+        static () => new DavDatabaseContext();
+
+    internal static Func<MetricsDbContext> MetricsContextFactory { get; set; } =
+        static () => new MetricsDbContext();
+
+    internal static Func<UsenetMigrationDbContext> UsenetMigrationContextFactory { get; set; } =
+        static () => new UsenetMigrationDbContext();
+
+    internal static Func<string> UsenetMigrationDatabaseFilePath { get; set; } =
+        static () => UsenetMigrationDbContext.DatabaseFilePath;
+
+    internal static Func<string> ContractFilePath { get; set; } =
+        static () => Path.Join(DavDatabaseContext.ConfigPath, ContractFileName);
+
+    private static readonly JsonSerializerOptions SerializerOptions = new()
+    {
+        PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+        WriteIndented = true,
+    };
+
+    /// <summary>
+    /// Writes the contract file. Never throws: the contract is informational and must
+    /// not block startup or the usenet-migration wizard.
+    /// </summary>
+    public static async Task WriteAsync(CancellationToken ct = default)
+    {
+        try
+        {
+            await WriteCoreAsync(ct).ConfigureAwait(false);
+        }
+        catch (Exception e) when (e is not OutOfMemoryException)
+        {
+            Log.Warning("Could not write database contract file to {Path}: {Reason}", ContractFilePath(), e.Message);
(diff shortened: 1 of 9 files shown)


hoivikaj:
Review findings addressed in `54833b37` and `1d5a7c35`:

1. **MainContextFactory provider branch (CodeRabbit, major)** — the default factory now creates `PostgresDavDatabaseContext` when `DATABASE_PROVIDER=postgres`, matching the explicit branch in `Program.cs`. Note: no wrong-data bug existed — `DavDatabaseContext.ConfigureOptions` already routes to Npgsql when postgres is configured, and `GetAppliedMigrationsAsync` reads the same `__EFMigrationsHistory` rows regardless of context type — but the explicit type removes the ambiguity. Covered by a new regression test in `PostgresMigrationTests` that runs in the CI postgres job: it asserts the default factory resolves to `PostgresDavDatabaseContext` and that a contract written against a migrated PostgreSQL schema reports `provider: "postgres"` with the applied history.
2. **Generic catch clause (code quality)** — the stale temp-file cleanup now catches `IOException` and `UnauthorizedAccessException` specifically.

Skipped: the CodeRabbit docstring-coverage warning (8.7% < 80%). The writer's class and entry point carry XML docs; per repo convention (AGENTS.md — comments only for non-obvious logic), tests and self-explanatory members stay undocumented.

hoivikaj:
Latest finding addressed in `86a995f7`:

- **Static seam race between test classes (CodeRabbit, minor)** — valid latent hazard: `DatabaseContractWriterTests` and `PostgresMigrationTests` are separate xUnit collections, so a full-suite run with `DATABASE_PROVIDER=postgres` could interleave their save/mutate/restore cycles on the writer's static seams. Both classes now live in a shared `DatabaseContractWriterCollection` with `DisableParallelization = true`, mirroring the existing `HttpIntegrationCollection` pattern for process-wide state. (CI jobs were not affected today: the postgres job filters to `PostgresMigrationTests` only, and the normal job skips the postgres test before it touches the seams.)

The earlier generic-catch finding was fixed in `54833b37` (typed `IOException`/`UnauthorizedAccessException` catches); the bot comment predates that push.

PR #1195, opened 2026-08-26
fix(sab): oversize NZBs and stalled indexer fetches are rejected before they fill the disk

### LINKED ISSUE #1182: NZB byte and time limits are enforced only after disk is consumed
## Summary

The 256 MiB decompressed NZB limit is checked only after the normalized XML has been fully copied to the blob store. Gzip expansion and `addurl` bodies can therefore consume disk beyond the accepted limit before rejection. For URL submissions, the 60-second timeout covers response acquisition through headers, not the later body copy.

Optional backup runs before validation, does not receive the request token, and is not removed when later validation rejects the NZB.

Found by the 2026-08-25 queue import audit at `3328f030`. Related: #1176 / #1177 (the accepted decompressed baseline is now 256 MiB; this issue is enforcing that limit *during* copy).

## Code evidence

- `NzbSubmissionService.SubmitAsync` calls `BlobStore.WriteBlob` before `NzbInputValidator.ValidateAndSumSegmentBytes`.
- Backup runs between those operations and `BackupNzbAsync` uses `CopyToAsync` without a cancellation token.
- `AddUrlRequest.GetAsync` uses `ResponseHeadersRead` and returns an open content stream after its local timeout source has been disposed.

## Proposed direction

Create one bounded ingestion primitive used by every submission adapter. Enforce decompressed bytes during copy, carry an end-to-end deadline and caller cancellation, and atomically clean temporary output. Validate before publishing backups.

## Acceptance criteria

- [ ] Decompressed input stops at the configured XML limit plus at most one bounded buffer.
- [ ] `addurl` header wait and body consumption share an end-to-end deadline.
- [ ] Oversized, gzip-bomb, slow-drip, cancelled, and malformed inputs leave no queue blob, temp blob, or backup.
- [ ] Valid NZBs up to 256 MiB continue to work through upload, URL, watch-folder, and migration entry points.

## Relevant files

- `backend/Queue/NzbSubmissionService.cs`
- `backend/Api/SabControllers/AddUrl/AddUrlRequest.cs`
- `backend/Utils/NzbStreamUtil.cs`
- `backend/Models/Nzb/NzbInputValidator.cs`
- `backend/Database/FileBlobStore.cs`

hoivikaj:
## Summary

Stacked on #1192 (diffs show only this change once it merges).

- The decompressed 256 MiB NZB limit is now enforced **during** the blob copy via a bounded read stream, so an oversize document or gzip bomb fails the copy with the usual 4xx validation error instead of writing unbounded data to disk first.
- Validation now runs **before** the NZB backup step; a rejected submission leaves neither a blob nor a backup file, and a failed/cancelled backup removes its own partial file.
- `addurl` now uses one shared 60 s deadline for the header wait **and** the response-body copy (previously the timeout source was disposed once headers arrived, so a slow-drip indexer could stall the ingest forever); hitting it returns a SAB-visible "download timed out" error.
- `LimitedReadStream` is promoted to `backend/Utils` with a caller-supplied limit exception; Warden call sites keep their existing behavior via `WardenInputLimits.CreateLimitedStream`.

Fixes #1182

## Test plan

- [x] New `LimitedReadStreamTests`: exact-limit copy succeeds, one byte over throws the factory exception, cancellation is not misreported as a limit trip.
- [x] New `NzbSubmissionIngestTests`: oversize plain XML fails during copy with no blob/temp file; gzip bomb trips the decompressed limit; backup-enabled validation failure leaves no backup file; valid NZB is still backed up.
- [x] New `AddUrlDeadlineTests`: slow-drip loopback server hits the shared deadline in ~0.5 s with a "timed out" 4xx and leaves no queue row or blob.
- [x] Existing `AddFileDuplicateReplaceTests`, `AddFileRequestTests`, Warden tests green.
- [ ] CI: full backend + frontend gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added consistent timeout handling when importing NZB files from URLs, including slow or stalled downloads.
  * Improved cancellation behavior when requests are interrupted.
  * Prevented incomplete temporary files, queue entries, and backups from remaining after rejected submissions.
  * Added decompressed-content limits to help protect against oversized or compressed NZB payloads.
  * Improved validation of total segment sizes before processing submissions.
* **Reliability**
  * Backup and restore operations now handle failures and cancellation more safely.
  * Added coverage for download deadlines, size limits, cleanup, and valid backup workflows.

### DIFF
--- a/backend/Api/SabControllers/AddUrl/AddUrlController.cs
+++ b/backend/Api/SabControllers/AddUrl/AddUrlController.cs
@@ -20,13 +20,26 @@ public class AddUrlController(
 {
     public async Task<AddUrlResponse> AddUrlAsync(AddUrlRequest request)
     {
+        // Owns the shared fetch/ingest deadline created in AddUrlRequest.New.
+        using var fetchDeadline = request.FetchDeadlineSource;
         var controller = new AddFileController(Context, dbClient, queueManager, Config, websocketManager);
-        var response = await controller.AddFileAsync(request).ConfigureAwait(false);
-        return new AddUrlResponse()
+        try
         {
-            Status = response.Status,
-            NzoIds = response.NzoIds,
-        };
+            var response = await controller.AddFileAsync(request).ConfigureAwait(false);
+            return new AddUrlResponse()
+            {
+                Status = response.Status,
+                NzoIds = response.NzoIds,
+            };
+        }
+        catch (OperationCanceledException ex) when (
+            fetchDeadline?.IsCancellationRequested == true && !Context.RequestAborted.IsCancellationRequested)
+        {
+            // The deadline fired while SubmitAsync was copying the response body;
+            // surface a fetch failure to the SAB client instead of an aborted request.
+            throw new BadHttpRequestException(
+                "Failed to fetch the nzb file: the download timed out.", ex);
+        }
     }
 
     protected override async Task<IActionResult> Handle()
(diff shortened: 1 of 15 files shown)


hoivikaj:
@coderabbitai review

PR #1197, opened 2026-08-26
fix(queue): busy database no longer marks completed downloads as failed

### LINKED ISSUE #1180: Transient database failures during finalization can fail or rerun successful imports
## Summary

The queue does not distinguish a content-processing failure from a transient database failure during a successful queue-to-history commit. A `DbUpdateException` from finalization falls into the generic processing-error handler, which immediately tries to finalize the same item as failed.

If that second write succeeds, otherwise healthy content is reported as a failed download. If it also fails, the original queue row remains eligible with no persistence backoff, so the coordinator can rerun the complete import.

Found by the 2026-08-25 queue import audit at `3328f030`. Not observed in the 2026-08-25 support dumps (those locks were on `metrics.sqlite`), but it is the same misclassification pattern and a latent trap on a write-heavy host.

Related: #622

## Code evidence

- `QueueItemProcessor.ProcessAsync` has dedicated handling for cancellation and retryable download errors, but not transient database errors from the successful completion call.
- The generic catch calls `MarkQueueItemCompleted(..., error: e.Message)`.
- `ExceptionExtensions.IsTransientDatabaseException` already identifies SQLite/PostgreSQL write-contention classes, but queue finalization does not use it.

## Failure sequence

1. Complete all NNTP and archive work successfully.
2. Inject SQLite `BUSY` or `LOCKED` during the completion save.
3. The generic catch attempts to create failed history for a database infrastructure error.
4. If the second save succeeds, the completed content is marked failed.
5. If the second save also fails, the persisted queue item remains immediately claimable and the expensive import can restart without delay.

## Proposed direction

Introduce an idempotent, bounded persistence-retry state separate from provider and content failure. Retry genuine contention with jittered backoff while preserving the successful processing result. Treat read-only, full-disk, corruption, and other operator-action conditions distinctly.

## Acceptance criteria

- [ ] SQLite `BUSY`/`LOCKED` cannot turn a successful import into failed history.
- [ ] Repeated persistence failures cannot hot-loop the complete NNTP pipeline.
- [ ] A recovered database commits one history row and one mount tree.
- [ ] Read-only, full-disk, corruption, and disk-I/O conditions produce one actionable operator-facing event without repetitive stack dumps.

## Relevant files

- `backend/Queue/QueueItemProcessor.cs`
- `backend/Queue/QueueManager.cs`
- `backend/Extensions/ExceptionExtensions.cs`
- `backend/Database/DavDatabaseContext.cs`

hoivikaj:
## Summary

Stacked on #1196 (diffs show only this change once it merges).

- The finalize commit now retries SQLITE_BUSY/LOCKED in place — 3 short attempts on the same staged ChangeTracker (blob writes are idempotent: same ids, temp-file + move) — so a contended commit no longer turns a completed import into a failed-history row.
- When contention persists, or the database reports disk-full/read-only/corruption, the item now **stays queued** with a `PauseUntil` backoff (~60 s transient, ~5 min disk/corrupt) and a single warning — instead of a misleading "download error" history entry or an immediately reclaimable row that re-downloads the whole NZB.
- A failed-history finalize that also hits contention now backs off the same way instead of being silently swallowed into a hot loop.
- SQLITE_IOERR and other unexpected codes keep the existing Error+stack behavior; nothing here retries disk/corrupt in place.

Fixes #1180

## Test plan

- [x] New `FinalizeDatabaseErrorTests` (full import through `QueueItemProcessor` with a save interceptor): one injected BUSY → in-place retry completes the import (exactly 2 finalize attempts); persistent BUSY → queued with ~60 s backoff, no history; codes 8/13/11 → no in-place retry, queued with ~5 min backoff; content failure + BUSY failed-finalize → paused instead of hot-looping.
- [x] Queue/history suites green (385 passed).
- [ ] CI: full backend + frontend gate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Queue items now remain queued when finalization encounters temporary database contention, disk failures, or corruption.
  * Added bounded retries and backoff to improve recovery from database errors.
  * Queue pause information and status updates are preserved more reliably.

* **Tests**
  * Added coverage for database finalization failures, retries, and backoff behavior.
  * Improved reusable test coverage for video segment processing.

### DIFF
--- a/backend/Api/SabControllers/AddUrl/AddUrlController.cs
+++ b/backend/Api/SabControllers/AddUrl/AddUrlController.cs
@@ -20,13 +20,26 @@ public class AddUrlController(
 {
     public async Task<AddUrlResponse> AddUrlAsync(AddUrlRequest request)
     {
+        // Owns the shared fetch/ingest deadline created in AddUrlRequest.New.
+        using var fetchDeadline = request.FetchDeadlineSource;
         var controller = new AddFileController(Context, dbClient, queueManager, Config, websocketManager);
-        var response = await controller.AddFileAsync(request).ConfigureAwait(false);
-        return new AddUrlResponse()
+        try
         {
-            Status = response.Status,
-            NzoIds = response.NzoIds,
-        };
+            var response = await controller.AddFileAsync(request).ConfigureAwait(false);
+            return new AddUrlResponse()
+            {
+                Status = response.Status,
+                NzoIds = response.NzoIds,
+            };
+        }
+        catch (OperationCanceledException ex) when (
+            fetchDeadline?.IsCancellationRequested == true && !Context.RequestAborted.IsCancellationRequested)
+        {
+            // The deadline fired while SubmitAsync was copying the response body;
+            // surface a fetch failure to the SAB client instead of an aborted request.
+            throw new BadHttpRequestException(
+                "Failed to fetch the nzb file: the download timed out.", ex);
+        }
     }
 
     protected override async Task<IActionResult> Handle()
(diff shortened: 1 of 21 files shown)


hoivikaj:
@coderabbitai review

PR #1297, opened 2026-09-01
fix(db): container no longer dumps a stack trace per file when local streaming metadata is corrupt

hoivikaj:
## Summary

A truncated/corrupt on-disk blob (unclean shutdown, partial `/config` restore, etc.) crashed health checks with an unhandled `MemoryPack.MemoryPackSerializationException` stack, once per affected file — a whole season sweeping together looked like dozens of "crashes" in the logs. The same failure would also 500 a WebDAV read instead of the existing clean 404, and could dump raw stacks in a few background repair/benchmark paths.

This PR treats an unreadable local blob as its own classified, handled failure — parallel to the existing `MissingFilePayloadException` handling, but with an accurate message (the payload is present but undecodable, not missing).

## What changed

- **`FileBlobStore.ReadBlob<T>`**: catches confirmed zstd/MemoryPack decode failures (`MemoryPackSerializationException`, `ZstdSharp.ZstdException`, `EndOfStreamException` for a truncated frame) and rethrows a new typed `CorruptedBlobPayloadException` carrying the blob id/path/payload type. Ordinary I/O, permission, and cancellation failures are untouched.
- **`DavDatabaseClient`**: the three typed getters (`GetDavNzbFileAsync`/`GetDavRarFileAsync`/`GetDavMultipartFileAsync`) now read through the injected `IBlobStore` and fall back to the legacy SQL row when the blob is unreadable — logging one handled Warning. If no legacy row exists, the exception is rethrown (never silently converted into a `null` "missing" result).
- **`ExceptionExtensions.TryGetKnownErrorMessage`**: classifies `CorruptedBlobPayloadException` early (before the general transport/download walk), so its actionable message isn't overwritten by an inner known exception like `EndOfStreamException`.
- **`HealthCheckService`**: a dedicated catch records `Unhealthy` + `ActionNeeded` with restore/re-download guidance, using the same daily→weekly recheck cadence as missing payloads, and never touches STAT/PAR2/Arr repair.
- **`ExceptionMiddleware`**: a dedicated catch returns 404 (reusing the existing `missing-file-payload` wire classification so the Explore player's terminal/no-retry state still applies) with a deduped, no-stack Warning.
- **Secondary readers** (`DavNzbFileBlobUpdater`, `BenchmarkCorpusProvider`, PAR2 zero-fill trigger, preflight pre-warm) recover via their existing fallback or log a handled Warning instead of an unhandled/undifferentiated stack.

## Test plan

- `FileBlobStoreTests`: empty, malformed-zstd, and truncated-zstd-frame blobs each become `CorruptedBlobPayloadException` with the right id/type/inner exception; a corrupted read is never cached and a later valid write recovers.
- `ExceptionExtensionsTests`: a wrapper whose inner exception is `EndOfStreamException` still surfaces the wrapper's actionable message.
- `DavDatabaseClientTests`: corrupt blob + legacy row falls back without throwing (Nzb/Rar/Multipart); corrupt blob with no legacy row rethrows `CorruptedBlobPayloadException`.
- `HealthCheckDegradedClassificationTests`: a corrupted blob with no legacy row records `Unhealthy`/`ActionNeeded`, logs exactly one Warning with no attached exception, no Error, and never contacts PAR2.
- `ExceptionMiddlewareTests`: typed 404 + header before response start, abort after start, one deduped no-stack Warning, and no streaming-failure/repair record.
- `DavNzbFileCorruptionRecordTests`: `DavNzbFileBlobUpdater.MutateAsync` recovers via the caller's fallback and writes a replacement blob when the current one is unreadable.

Per `AGENTS.md`, PR CI covers backend build/analyzers/tests — not re-run locally.

## Notes

- No database migration or docs page needed.
- `StreamingPayloadExistsAsync` / `RemoveMissingPayloadsTask` are intentionally unchanged (existence-only checks) so this never causes automatic deletion of recoverable data.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of unreadable streaming metadata.
  * Automatically uses legacy data when available.
  * Reports unrecoverable payloads as actionable 404 errors with recovery guidance.
  * Prevents corrupted data from being cached and allows repair workflows to continue safely.
  * Health checks defer unreadable payloads without removing or blocklisting affected items.

* **Tests**
  * Added coverage for corrupted, truncated, and invalid payloads across retrieval, error handling, health checks, and repair workflows.

### DIFF
--- a/backend/Database/DavDatabaseClient.cs
+++ b/backend/Database/DavDatabaseClient.cs
@@ -220,6 +220,16 @@ public sealed class DavDatabaseClient(DavDatabaseContext ctx, IBlobStore? blobSt
     /// the exception is rethrown so callers see the real corruption instead of a false
     /// "missing payload" result.
     /// </summary>
+    /// <summary>
+    /// Reads a typed blob from the blob store, recovering from a legacy database row
+    /// if the blob is corrupted and a fallback is available.
+    /// </summary>
+    /// <typeparam name="T">The deserialized payload type.</typeparam>
+    /// <param name="davItem">The DAV item owning this blob.</param>
+    /// <param name="blobId">The blob identifier to read.</param>
+    /// <param name="readLegacyRow">A callback to query the legacy database row when the blob is corrupted.</param>
+    /// <param name="ct">Cancellation token.</param>
+    /// <returns>The deserialized payload, fallback row, or throws if the blob is corrupted and no fallback exists.</returns>
     private async Task<T?> ReadTypedBlobOrFallbackAsync<T>(
         DavItem davItem,
         Guid blobId,
--- a/backend/Database/DavDatabaseClient.cs
+++ b/backend/Database/DavDatabaseClient.cs
@@ -2,6 +2,7 @@ using System.Text;
 using System.Runtime.CompilerServices;
 using Microsoft.EntityFrameworkCore;
 using NzbWebDAV.Database.Models;
+using NzbWebDAV.Exceptions;
 using NzbWebDAV.Extensions;
 using NzbWebDAV.Services;
 using Serilog;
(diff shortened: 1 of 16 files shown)


hoivikaj:
@coderabbitai full review

PR #1310, opened 2026-09-02
fix(webdav): directory listings no longer fail with HTTP 500 when a filename contains XML-invalid characters

### LINKED ISSUE #1308: WebDAV PROPFIND returns HTTP 500 when a filename contains U+FFFE (invalid XML char) — whole directory unlistable
## Summary

WebDAV directory listing (PROPFIND) fails with **HTTP 500** when any child filename stored in the library contains code points that are invalid in XML 1.0 — specifically **U+FFFE**. Because the whole listing request fails, every other file in that directory becomes unreachable through WebDAV until the offending entry is removed.

## Impact chain

This bricks directory traversal for any client that needs a full listing:

- rclone-mounted WebDAV surfaces it as `IO error: couldn't list files: 500 Internal Server Error`, and the mount reports `Input/output error` for the directory.
- Tools that import from the mounted tree (e.g. Sonarr/Radarr-style importers) cannot read the folder at all, so downloads that completed successfully never finish importing.
- One bad filename poisons the entire containing directory — all healthy sibling files in it are collateral damage.

## Root cause

The exception is thrown while serializing the WebDAV XML response, before any of it is sent:

```
System.ArgumentException: '￾', hexadecimal value 0xFFFE, is an invalid character.
   at NzbWebDAV.Middlewares.ExceptionMiddleware.InvokeAsync(HttpContext context)
```

XML 1.0 forbids U+FFFE (and U+FFFF, plus most C0/C1 control chars). The stored filename contains one, the XML writer throws, and the middleware converts it to a generic 500.

## How names get corrupted

The affected names originate from NZBs whose file entries were encoded in a legacy charset (the pattern is typical of Latin-1/French-accented scene releases that were misdecoded). Characters that should be `ñ`, `é`, `è`, `à`, `û` arrive as a mangled sequence. Observed examples (escaped, actual stored names):

- `Mucha.Lucha.S02E20.Poocha.Lucha.Cinco.De.Pi\uFFFE\uE0C3\uE0B1ata.1080p.AMZN.WEB-DL.DDP2.0.H.264-NTb.mkv` (should be `Piñata`)
- `Mucha.Lucha.S02E15.El.Ni\uFFFE\uE0C3\uE0B1o.Loco.La.Bruja.1080p.AMZN.WEB-DL.DDP2.0.H.264-NTb.mkv` (should be `Niño`)
- `The.Loud.House_S07E03-E04_Child\uFFFE\uE0C3\uE0A2\uE0C2\uE080\uE0C2\uE099s.Play.Force.of.Habits.mkv` (should be `Child's Play`)
- `1073 One Piece S21E182 Multi ''... Onigashima, enfer \uFFFE\uE0C3\uE0A0 ciel ouvert !'' ... WebDl1080p ! 2023.mkv` (should be `à ciel ouvert`)

i.e. an invalid-character marker (U+FFFE) followed by private-use-area code units that encode the original UTF-8 bytes.

## Steps to reproduce

1. Have a download item whose stored filename contains U+FFFE (or U+FFFF, or a control char like U+0001) — the mojibake above is one way, but even a hand-crafted entry suffices.
2. Send `PROPFIND` with `Depth: 1` on the parent collection.
3. Server responds `500 Internal Server Error`; logs show the `ArgumentException` above.

Expected: the listing succeeds; entries with XML-invalid characters are sanitized (stripped or replaced, e.g. `\uFFFD` or an `_xFFFF_`-style escape) rather than failing the entire request. Optionally, names should be sanitized at ingest so invalid code points never enter storage in the first place.

## Environment

- InfiniDysk image built 2026-08-28 (digest `sha256:8cf616c1...`)
- WebDAV access via rclone with `--links` (mount + `*arr`-style importer), and via direct PROPFIND

hoivikaj:
## Summary

Fixes #1308.

A single stored filename containing a character XML 1.0 forbids (U+FFFE, U+FFFF, lone surrogates, C0 controls) made `XmlWriter` throw while serializing the PROPFIND multistatus body. The whole directory listing became a 500 — rclone surfaced it as `IO error: couldn't list files`, and importers could not read healthy sibling files.

- **Response-time fix:** `displayname` for items and collections now runs through a new `XmlTextUtil.ReplaceInvalidXmlChars` (replaces with U+FFFD), so existing corrupted entries list correctly without a DB migration. `href` was already safe (percent-encoded via `Uri.AbsoluteUri`).
- **Ingest fix:** `PathSanitizer.SanitizeComponent` strips XML-invalid characters (as `_`) in both Windows-safe and minimal modes, so mis-decoded NZB subjects (e.g. `Pi\uFFFE\uE0C3\uE0B1ata`) never enter storage with them again.

## Test plan

- [x] New `XmlTextUtilTests` (valid text untouched, U+FFFE/U+FFFF/C0/lone surrogates replaced, surrogate pairs preserved)
- [x] New `DisplayNamePropertyTests` — property managers return a sanitized `displayname` that serializes through `XmlWriter`
- [x] `PathSanitizerTests` extended for both modes
- [x] Existing `PropFindHandlerPatchTests` / `WebDavContractTests` still pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid XML characters in sanitized paths and WebDAV item and collection display names.
  * Prevented malformed display names from causing XML serialization issues.
  * Preserved valid Unicode characters, including valid surrogate pairs.

* **Tests**
  * Added coverage for invalid characters, unpaired surrogates, custom replacements, and successful XML serialization.

### DIFF
--- a/backend/Utils/PathSanitizer.cs
+++ b/backend/Utils/PathSanitizer.cs
@@ -35,7 +35,7 @@ public static class PathSanitizer
             return SanitizeMinimal(name);
 
         var sb = new StringBuilder(name.Length);
-        foreach (var ch in name)
+        foreach (var ch in XmlTextUtil.ReplaceInvalidXmlChars(name, '_'))
         {
             if (ch < 0x20 || WindowsInvalidChars.Contains(ch))
                 sb.Append('_');
@@ -67,13 +67,14 @@ public static class PathSanitizer
     }
 
     /// <summary>
-    /// Minimal sanitization when Windows-safe paths are disabled: only '/' and NUL.
+    /// Minimal sanitization when Windows-safe paths are disabled: only '/', NUL, and
+    /// characters XML 1.0 forbids (they would break every WebDAV listing of the parent).
     /// Still length-capped so a sanitized name can never overflow DavItem.Name (255).
     /// </summary>
     private static string SanitizeMinimal(string name)
     {
         var sb = new StringBuilder(name.Length);
-        foreach (var ch in name)
+        foreach (var ch in XmlTextUtil.ReplaceInvalidXmlChars(name, '_'))
         {
             if (ch is '/' or '\0')
                 sb.Append('_');
--- a/backend/Utils/XmlTextUtil.cs
+++ b/backend/Utils/XmlTextUtil.cs
@@ -1,0 +1,60 @@
+using System.Xml;
+
+namespace NzbWebDAV.Utils;
+
+/// <summary>
+/// XML 1.0 forbids U+FFFE, U+FFFF, lone surrogates, and most C0 controls. Names
+/// carrying those (e.g. mis-decoded NZB subjects) make XmlWriter throw while
+/// serializing a PROPFIND response, which turns the whole listing into a 500.
+/// </summary>
+public static class XmlTextUtil
+{
+    public const char ReplacementChar = '\uFFFD';
+
+    public static bool ContainsInvalidXmlChars(string text)
+    {
+        for (var i = 0; i < text.Length; i++)
+        {
+            if (XmlConvert.IsXmlChar(text[i]))
+                continue;
+
+            if (i + 1 < text.Length && XmlConvert.IsXmlSurrogatePair(text[i + 1], text[i]))
+            {
+                i++;
+                continue;
+            }
+
+            return true;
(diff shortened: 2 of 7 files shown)


TheMrClaus:
💪

PR #1317, opened 2026-09-03
chore(ci): frontend mutation runs no longer take 30+ minutes on dependency-only PRs

hoivikaj:
## Summary

The frontend mutation job in #1316 (a `qs` lockfile bump) ran a full, cold StrykerJS pass. Two causes:

- `frontend/package-lock.json` was in the trigger paths, so every Dependabot npm bump ran mutation testing over the whole `app/clients` + `app/auth` pilot (~1.7k LOC) even though no pilot source changed.
- The incremental cache key was per-`ref_name` (e.g. `1312/merge`) and PR-scoped caches are not readable from other PRs, so each PR started cold. Because the key had no commit component, the cache was also never refreshed after its first save.

Changes:

- Drop `package-lock.json` from the frontend pilot-change detection (regular CI already covers dependency bumps).
- Add a `push` to `main` trigger (frontend pilot paths only) that seeds a `stryker-js-<config-hash>-main-*` incremental cache, and add it as a fallback `restore-key` for PRs. Key now includes `github.sha` so the cache is refreshed on each run.
- Run StrykerJS with `--concurrency 3` on the 4-vCPU runner instead of the config default of `50%` (=2).

The backend job is unchanged on PRs and does not run on the new `push` trigger.

## Test plan

- [x] `mutation.yml` parses as YAML
- [ ] Dependabot npm-only PR: frontend mutation job reports "No frontend mutation-pilot files changed" and skips
- [ ] PR touching `frontend/app/clients/**`: job restores the `-main-` cache once one has been seeded, and total runtime drops

## Summary by CodeRabbit

- **Chores**
  - Improved automated quality checks for backend and frontend changes.
  - Updated mutation-testing triggers for pull requests, pushes to the main branch, and supported manual workflows.
  - Refined change detection and manual workflow scope options.
  - Improved test result reuse across branches, configurations, and commits.
  - Increased frontend mutation-testing parallelism to complete validation more efficiently.

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -8,13 +8,13 @@ on:
       - frontend/app/clients/**
       - frontend/app/auth/**
       - frontend/stryker.config.json
+      - frontend/package-lock.json
   workflow_dispatch:
     inputs:
       scope:
         description: Mutation scope
         type: choice
         options:
-          - changed
           - queue
           - streams
           - usenet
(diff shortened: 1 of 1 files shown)


hoivikaj:
Addressed both review comments in the latest commit:
- Removed `changed` from the `workflow_dispatch` scope options (and from both job `if` lists); PR events still use the changed-files selector.
- Added `frontend/package-lock.json` back into the Stryker incremental cache key and both restore keys, and added it to the push-to-main paths so the shared `-main-` cache is re-seeded after dependency changes.

PR #1319, opened 2026-09-03
feat(ui): warn when Segment Cache duplicates rclone VFS read-ahead

hoivikaj:
## Summary

Users running a **Symlinks** library stream through an rclone mount. When that mount uses `--vfs-read-ahead`, rclone already buffers ahead of playback, so InfiniDysk's Segment Cache only adds disk writes. InfiniDysk cannot always inspect the mount, so this PR adds guidance instead of silently changing behavior:

- **Streaming tab** — a dismissable warning appears when Segment Cache is on and `api.import-strategy` is `symlinks`. Dismissal is remembered per browser (`localStorage`).
- **Rclone tab** — when the RC connection test can read VFS statistics and confirms read-ahead is enabled while Segment Cache is on, a definitive (non-dismissable, per-test) warning is shown.
- Both warnings link to the **Setup Guide** with a `returnTo`, and the copy encourages existing installations to re-run it since it reviews recommended configuration.
- Docs: new "Segment Cache and rclone read-ahead" section in `docs/configuration/streaming.md`; Setup Guide intro now notes it is worth completing on long-running installs.

No config keys or backend changes; setup wizard version is unchanged (advisory-only UI).

## Test plan

- [x] `vitest` — new tests cover the warning predicate, the dismiss button persisting to `localStorage` and staying hidden on remount, and no warning for STRM libraries
- [x] ESLint (zero warnings), Prettier, `tsc -b` on the frontend
- [ ] Manual: Settings → Streaming with a Symlinks library and Segment Cache on shows the notice; dismiss survives reload
- [ ] Manual: Settings → Rclone → Test Conn against a mount with `--vfs-read-ahead` shows the read-ahead warning when Segment Cache is on

## Summary by CodeRabbit

- **New Features**
  - Added warnings in Streaming and Rclone settings when Segment Cache and rclone read-ahead are enabled together for Symlink libraries.
  - Warnings link to Streaming settings and the Setup Guide and can be dismissed with the choice remembered.
  - Warnings are hidden for STRM libraries and during initial page loading.
  - Rclone connection results now reflect the detected read-ahead configuration.

- **Documentation**
  - Added guidance on configuring Segment Cache and rclone read-ahead.
  - Updated the setup guide with recommendations for existing installations and approved settings.

### DIFF
--- a/docs/configuration/streaming.md
+++ b/docs/configuration/streaming.md
@@ -65,6 +65,21 @@ path) is local SSD/NVMe or other storage that can safely absorb the extra writes
 Disable the cache for slow disks, network mounts, or flash storage with limited
 write endurance; alternatively, point **Cache path** at suitable local storage.
 
+### Segment Cache and rclone read-ahead
+
+Symlink libraries stream through an rclone mount. When that mount runs with
+`--vfs-read-ahead`, rclone already buffers ahead of playback, and Segment Cache only
+adds disk writes without improving seeks. Keep one or the other: the
+[Setup Guide](../getting-started/setup-guide.md) disables Segment Cache for Symlinks
+and recommends the bounded rclone VFS cache instead.
+
+InfiniDysk cannot always inspect the mount, so the Streaming tab shows a dismissable
+warning whenever Segment Cache is on for a Symlinks library. Dismissal is remembered
+per browser. When the [rclone RC connection test](rclone.md) can read VFS statistics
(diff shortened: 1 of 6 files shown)


hoivikaj:
Addressed the CodeRabbit finding: connection tests now track a request generation that is bumped whenever host/user/password change, so a stale in-flight response can no longer restore `readAheadBytes` or show the Segment Cache warning for an untested configuration. Added `rclone.test.tsx` with a regression test covering an input change while a test is pending.

PR #1340, opened 2026-09-04
feat(ui): activity chart separates client playback reads from app-triggered reads

hoivikaj:
## Summary
- Split Overview Activity article reads into solid client playback and dashed app-triggered series.
- Persist token-context workload attribution through raw fetches, rollups, retention, provider remapping, and Overview API totals.
- Historical pre-upgrade rows default to Unknown and display as app activity; no historical backfill is performed.

## Validation
- `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --filter "FullyQualifiedName~Metrics|FullyQualifiedName~GetOverviewStats|FullyQualifiedName~MultiProviderNntpClient|FullyQualifiedName~SegmentCacheNntpClient|FullyQualifiedName~StartupDatabaseMigration"`
- `dotnet test tests/NzbWebDAV.ArchitectureTests/NzbWebDAV.ArchitectureTests.csproj -c Release`
- `cd frontend && npx vitest run app/routes/overview && npm run typecheck && npm run lint`
- `./scripts/export-admin-openapi.sh`

## Upgrade notes
- Back up `/config` before upgrading: the additive metrics SQLite migration runs automatically at startup.
- No setup-wizard change is required; this feature adds no configuration.
- Manual private/shared-stream and queue/maintenance attribution checks require non-production Usenet credentials and were not run locally.

## Summary by CodeRabbit

* **New Features**
  * Overview statistics now distinguish client article reads from application article reads.
  * Throughput charts display separate client and application article series with updated legends, tooltips, and accessibility descriptions.
  * Client article totals are tracked across minute, hourly, and lifetime metrics.
  * Fetch activity is categorized by workload for more detailed reporting.
* **Bug Fixes**
  * Improved consistency of article-read aggregation across reporting windows and providers, including rerun and late-arriving data.
* **Tests**
  * Added coverage for chart rendering, aggregation, workload classification, and metric persistence.

PR #1350, opened 2026-09-07
fix(metrics): rclone/Plex playback through shared streams no longer shows up as "App reads" on the Activity chart

hoivikaj:
## Summary

With only rclone/Plex reading (no health checks or queue work running), the Overview **Activity** chart attributed most article fetches to **App reads** instead of **Client reads**.

Root cause: #1340 classifies a fetch as client/streaming only when the `AsyncLocal` read-session id is present. The shared-stream pump (`SharedStreamEntry.StartPump`) deliberately starts under `ExecutionContext.SuppressFlow()` so it never inherits a single reader's request state — which also drops the read-session id. Every article the pump fetched for attached clients was therefore recorded as `Background` and rendered as "App reads". Large open-ended rclone GETs go through the shared pump, while small ranges use a private stream (with a session) — hence the ~4k client / ~27k app split.

Fix: mark the pump's async flow with a new attribution-only `SharedStreamPumpContext`, and have `DownloadWorkloadClassifier.ClassifyForMetrics` treat it as `Streaming`. Admission priority and per-session diagnostics are unchanged.

## Test plan

- [x] `DownloadWorkloadClassifierTests`: pump marker → `Streaming`; disposing the marker restores `Background`.
- [x] `SharedStreamEntryTests.Pump_ClassifiesUpstreamReadsAsClientStreaming_WithoutInheritingReaderSession`: upstream reads issued by the pump see no read-session id (flow still suppressed) and classify as `Streaming`. Verified the test fails with the marker removed.
- [x] `dotnet test` filtered to classifier, shared-stream entry/registry, and terminal-miss logging suites: 52/52 pass.
- [x] `dotnet format --verify-no-changes` on touched files.
- [ ] Manual: with rclone/Plex scanning, Activity chart "Client reads" should now carry the bulk of the reads.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved workload metrics so shared-stream activity is correctly classified as streaming, even without an active read session.
  - Upstream reads from shared streams are now attributed to client streaming without inheriting the caller’s read-session identifier.

- **Tests**
  - Added coverage confirming streaming classification and proper restoration of workload context after shared-stream operations.

### DIFF
--- a/backend/Clients/Usenet/Contexts/DownloadWorkloadClassifier.cs
+++ b/backend/Clients/Usenet/Contexts/DownloadWorkloadClassifier.cs
@@ -20,7 +20,7 @@ internal static class DownloadWorkloadClassifier
 
     internal static SegmentFetch.FetchWorkload ClassifyForMetrics(CancellationToken cancellationToken)
     {
-        if (MultiProviderNntpClient.CurrentReadSessionId is not null)
+        if (MultiProviderNntpClient.CurrentReadSessionId is not null || SharedStreamPumpContext.IsActive)
             return SegmentFetch.FetchWorkload.Streaming;
 
         return Classify(cancellationToken) switch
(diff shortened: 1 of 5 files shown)


hoivikaj:
@coderabbitai full review

PR #1362, opened 2026-09-08
feat(rclone): built-in rclone mount managed from the admin UI

### LINKED ISSUE #254: Built-in rclone mount managed from the admin UI
### Before submitting

- [x] I have searched existing issues and did not find a similar request.
- [x] This request relates to use with legally obtained content.

### Problem or use case

NzbDav exposes a WebDAV server, but to turn that into a usable filesystem for Sonarr/Radarr symlink imports and Plex/Emby/Jellyfin playback, users have to run a separate rclone process or sidecar container. That means maintaining a second copy of the WebDAV URL and credentials, hand-writing rclone.conf, and tuning VFS cache flags by hand — a recurring source of misconfiguration (content-length mismatches, wrong VFS cache mode, mounts that drift out of sync when WebDAV credentials change). Altmount already ships rclone in-image and mounts its own WebDAV internally with zero external dependency. Parity here would remove a whole class of setup friction and support questions, and make NzbDav genuinely single-container.

### Proposed solution

- Bundle the rclone binary in the Docker image.
- Add an rclone mount section to the admin UI that manages an internal mount of NzbDav's own WebDAV:
        - enable/disable toggle
        - mount path
        - VFS cache mode + cache settings (size, max-age, poll-interval, etc.)
        - passthrough for advanced/custom flags
- NzbDav generates the rclone remote automatically from its existing WebDAV URL and credentials — no manual rclone.conf, no duplicate secret to keep in sync.
- Manage the mount lifecycle with the container (start/stop alongside the app) and surface mount health/status in the UI.
- Provide an explicit "use external rclone" mode that disables the integrated mount, so users who already run their own tuned mount aren't forced onto the built-in one. This keeps existing deployments working unchanged.

### Alternatives considered

- Keep running a separate rclone container/sidecar (today's approach) — works, but duplicates credential management and pushes all VFS tuning onto the user.
- Ship only a documented "recommended rclone config" — lowers the bar but doesn't remove the external dependency or the misconfiguration surface.

### Relevant area

WebDAV or streaming

### Additional context

Altmount's integrated rclone is good prior art for the config surface and lifecycle handling. One thing worth designing for up front: some setups benefit from more than one mount profile against the same WebDAV — e.g. a full-VFS-cache mount for Plex playback and a cache-off mount for Sonarr/Radarr analysis. Even if multi-mount is out of scope for v1, keeping the settings schema open to named mount profiles would avoid a breaking change later.

---

samtheruby:
Closes https://github.com/infinidysk/infinidysk/issues/254

## 💪 What

- Adds a built-in rclone mount: InfiniDysk runs `rclone rcd` itself and manages mounts from **Settings → Rclone**, so one container can expose the WebDAV tree as a folder for Sonarr, Radarr, Plex, Emby, and Jellyfin.
- Offers it in the setup wizard. The symlinks step now asks where rclone runs — inside InfiniDysk, or in your own container — and the built-in path is the default for a new install. Choosing it hides the sidecar flags, RC host, and read-ahead confirmation, because none of them describe anything that exists.
- Supervises the mounts as well as the process: mounts are applied when the daemon starts and whenever the mount list changes, so a container restart brings the library back without anyone opening the settings page.
- Adds an import that moves an existing rclone container into built-in mode, reading its live mount settings over the RC API — path, cache mode, tuning, and `--links` — with a pasted `rclone mount ...` command as the fallback when RC is not enabled.
- Bundles rclone and `fuse3` in the image at a pinned version, verifying the release's `SHA256SUMS` against rclone's signing key before trusting any checksum in it.
- Adds five config keys — `rclone.builtin.enabled`, `.mounts`, `.rc-port`, `.cache-dir`, `.cache-size-limit` — reachable through the existing `NZBDAV_CONFIG__` namespace.
- Adds `/api/rclone-mounts/{status,apply,remount,logs,import-external,webdav-credentials,clear-cache}` and regenerates the admin OpenAPI contract. The compatibility check reports no removed or renamed operations.
- Adds a **Clear Rclone Cache** maintenance task: releases the mounts, deletes rclone's own cache subtrees, and puts the mounts back.
- Detects and explains the environmental failures rather than leaving them to be discovered: a container without `--device /dev/fuse` is told which flags to add, a mount that misses `rclone.mount-dir` is flagged as breaking imported files, and a mount point inside `CONFIG_PATH` — or over `/app`, `/usr`, or another system directory — is rejected.
- Updates `docs/configuration/rclone.md`, adds a sidecar migration runbook to `docs/guides/mounting-webdav.md`, and adds a README entry.

Built-in mode is off by default for existing installations. One already running its own rclone container sees no new process, no new logs, and no behavior change, and its external Rclone Server settings are never modified. The setup wizard's version is unchanged, so nobody is re-prompted. No database migrations.

## 🤔 Why

- Running rclone separately is a recurring source of misconfiguration: the WebDAV credentials are duplicated by hand, the VFS flags drift from what the docs recommend, and `rclone.mount-dir` and the actual mount point are maintained in two places with nothing checking they agree.
- Getting the mount point wrong is silent and expensive. Imported files resolve through `<rclone.mount-dir>/.ids/...`, so a mount anywhere else leaves an entire library unplayable while every mount on screen looks healthy. That value already exists in config, so the mount path is derived from it and a mismatch is called out.
- The migration is the part that has to be right, since most people arrive with a tuned sidecar and a library of symlinks depending on it. Reading the running daemon rather than a config file means drift between a Compose file and reality cannot survive the import, and the sidecar is never modified, so rolling back is turning the feature off and starting it again.
- Single-container deployment brings InfiniDysk to parity with comparable tools, and the mount schema takes multiple mounts so different cache profiles can coexist later.

## 🧹 Mounts outlive the process that made them

This is the failure mode that shaped most of the lifecycle work here, and it is worth stating plainly: a FUSE mount survives the death of the process that created it. A container killed rather than stopped leaves its mount in the kernel's table, and the next start cannot mount over it — rclone refuses with *directory already mounted* and the library stays offline until someone clears it by hand.

Every end is handled:

- **On shutdown**, the mounts are released over the RC API before the process is stopped, rather than relying on rclone reacting to a signal in time. Two things make that reliable. The daemon is registered last among the hosted services, so it stops *first* rather than after thirty others have spent the shared `HostOptions.ShutdownTimeout`; and its unmount budget is independent of the host's token, because a token already cancelled by whatever stopped ahead of it means the request never leaves the process.
- **On startup, and after an unexpected daemon exit**, a mount left on one of our own configured mount points is cleared — but only when the kernel reports a `fuse` filesystem mounted exactly there *and* that mount no longer answers (`ENOTCONN`). A mount that still responds is reported and left alone, because something is serving it, usually an external rclone that has not been stopped yet, and taking it down would break a library mid-playback. The probe is bounded: a mount that never answers is treated as alive, since a read that hangs is not evidence of a dead mount and blocking the supervisor is the worse outcome.
- **On a settings change**, the daemon stops and the replacement starts on the next pass rather than immediately, so anything the old process left behind is swept while nothing is running. Restarting in the same pass put a healthy daemon in front of every later sweep, which meant a leftover could never be cleared.
- **Before mounting**, the supervisor waits for the freshly launched daemon to open its remote-control port. Mounting into that window produced a *connection refused* warning that read like a broken deployment rather than a race with startup.
- **Mount changes are serialized.** The supervisor's poll and the three admin endpoints all reconcile the same daemon; two passes at once both read "nothing is mounted" and both mount, and the loser reports *already mounted*.

## 💾 Cache: automatic by default, settable when you want

rclone's own `--vfs-cache-max-size` default is "off", which means a `full` cache mode mount grows until the filesystem is full, and the default cache directory sits beside the databases. Left alone, the ceiling is derived from the free space on that volume — half of it, capped at 20 GB, never below 256 MB — reserving 5 GB when the cache shares a volume with the databases, so a full cache cannot take the database down with it.

The cache directory and a **cache size limit** are both editable in the tab, always visible rather than behind a disclosure, since between them they decide how much disk rclone takes. Leaving the limit empty restores the automatic sizing. The tab shows the directory, its free space, and the ceiling in force.

## 🎛️ Defaults match the sidecar configuration we document

The built-in mount previously ran on defaults quite different from the flags this project recommends for a sidecar. They now agree:

| Setting | Was | Now |
|---|---|---|
| `--dir-cache-time` | 20 seconds | 1 week |
| `--vfs-cache-max-age` | 24 hours | 1 week |
| `--vfs-read-ahead` | unset | 512 MiB |
| `--buffer-size` | 16 MiB per transfer | 0 |
| `--poll-interval` | 60 seconds | 0 |
| `--use-cookies` | off | on |

A week-long directory cache is only safe because InfiniDysk drops a folder from that cache itself whenever it adds or removes an item, and a failed drop is retried and then logged rather than swallowed. That invalidation now reaches the built-in daemon — it previously went only to the external rclone, behind a setting that defaults to off, so on a built-in install it was doing nothing at all. **Remember folder listings for** is editable per share for anyone who wants it shorter.

`--buffer-size=0` matters for memory: rclone holds that much per in-flight transfer, which multiplies across concurrent streams, and the read-ahead buffers on disk instead. `--allow-non-empty` is deliberately *not* adopted — it hides real files under a mount and papers over the leftover-mount problem the shutdown and sweep paths now solve properly — and it has been removed from the flags the wizard hands out.

## 👀 Usage

The container needs FUSE access and shared bind propagation before built-in mode can work:

```yaml
  nzbdav:
    image: ghcr.io/infinidysk/infinidysk:latest
    devices:
      - /dev/fuse:/dev/fuse:rwm
    cap_add:
      - SYS_ADMIN
    security_opt:
      - apparmor:unconfined
    volumes:
      - ./config:/config
      - /mnt:/mnt:rshared
```

`rshared` is what decides whether anything outside the container can see the mount. Without it Plex and the Arr apps see an empty folder while InfiniDysk correctly reports the mount as healthy, because from inside the container it is.

On a new install, the setup wizard offers it directly. On an existing one, go to **Settings → Rclone** and switch on *Run rclone inside InfiniDysk*. Either press **Read my rclone server** to import an existing instance, or enter the WebDAV password once and add a mount at the offered path.

## 👩‍🔬 How to validate

**A fresh install, through the wizard**

1. Start the container with the flags above and run the setup wizard.
2. On the library step choose **Symlinks**. The playback step should offer *Run rclone inside InfiniDysk*, selected, with no sidecar flags, RC host, or read-ahead checkbox on screen.
3. Finish setup, then open **Settings → Rclone**. Built-in mode is on with one mount at the directory the wizard asked for, and the tab asks for the WebDAV password.
4. Enter it. The mount appears.

**A fresh install, from Settings**

1. Switch on *Run rclone inside InfiniDysk*. The daemon starts; because no remote exists yet, the tab asks for the WebDAV password.
2. Enter a deliberately wrong password. It is refused with a message naming where to find the right one, and no remote is created.
3. Enter the correct one. The mount appears, and the offered path matches `rclone.mount-dir` rather than a generic placeholder.
4. `ls` the mount point on the host and from another container. Both should list `.ids`, `completed-symlinks`, `content`, `nzbs`. The second proves bind propagation.
5. Play something imported through symlinks. It should resolve and stream.

**Migrating from an existing rclone container**

1. With an rclone sidecar running and its RC settings filled in under Server connection, press **Read my rclone server**.
2. Check the preview reproduces your mount path, cache mode, and `--links` setting, and that it warns you to stop the sidecar. Confirm your sidecar is untouched. Do not save yet.
3. Stop the sidecar, then press **Use these settings** and save. Saving applies the mount list straight away, which is why the sidecar has to stop first.
4. Enter the WebDAV password once under *Connect to your library*. The import deliberately does not copy it.
5. Confirm previously imported files still play — the mount path is preserved so their symlinks keep resolving.

**The blocked path**

Restart the container without `--device /dev/fuse` and open the tab. It should stop describing settings and print the compose changes to make, rather than failing opaquely or crash-looping.

**A mount left by a hard shutdown**

`docker kill` the container while a mount is up, then start it again. The leftover mount should be cleared on the way in rather than blocking the new one. Start an external rclone on the same path instead and it should be reported and left alone.

**Reclaiming cache disk**

Lower the cache size limit, then run **Settings → Maintenance → Clear Rclone Cache**. It should report the space freed and remount; playback stalls while it runs and resumes afterwards.

**Rolling back**

Switch built-in mode off and start the sidecar again. Nothing about the external instance was modified and the saved Rclone Server settings were never overwritten.

## 🔎 Four review rounds since the first draft

Every round was verified against the code before anything was changed, and findings that were already fixed, or wrong, were skipped with a reason rather than churned. The ones that changed behaviour, grouped by what they protected:

**Data loss.** Clear Cache deleted *every* child of the cache directory. That directory is an operator-set absolute path, so pointing it at `CONFIG_PATH` or a media root turned the button into a delete of databases and libraries. It now removes only `vfs/` and `vfsMeta/` — verified against rclone v1.75.1, which creates those two and nothing else — skips a link standing where one of them should be, and decides whether a daemon is running *inside* the mount gate so a daemon that starts mid-request cannot have its cache pulled out from under it.

**Supply chain.** The installer verified the clearsigned `SHA256SUMS` and then selected the checksum from the original file with `awk`. Appending an unsigned entry to a genuine older release produced exit code 0, the expected `VALIDSIG`, and exactly one match — the unsigned one. The checksum is now taken only from GnuPG's authenticated output, with the signer confirmed through the status stream and duplicate entries refused.

**Mounts that stay down.** After one successful pass over an unchanged mount list, the supervisor returned early forever. A manual remount whose replacement failed left the library down until someone clicked again or restarted. Endpoint-driven reconciles now clear that record, and a five-minute recheck covers a mount that disappears while `rcd` stays alive.

**Invalidation reaching nothing.** Starting the built-in daemon made it the exclusive `vfs/forget` target even before it had a working mount, silently cutting off a sidecar that was still mounted and being read. Both are notified now. Separately, rclone refuses an unqualified `vfs/forget` once more than one VFS is active, and it joins whatever path it is given onto the VFS root without reporting a miss — so a subtree mount was silently forgetting `/content/content/…` while the directory that changed stayed cached. Paths are translated per VFS and the request is qualified whenever more than one mount is up.

**Settings that did not apply, or applied themselves.** Changed VFS tuning was compared against live `vfs/stats` rather than assumed; an unedited wizard mount no longer reads as dirty because the frontend now mirrors the backend's read-ahead default; and focusing and leaving the read-ahead or cache-size box no longer rewrites the value, because a rounded display re-parsed is not an edit.

**Setup.** Re-running the wizard replaced the saved mount list with one derived mount, discarding extra shares and tuning. It now keeps what it finds. The built-in and external branches also disagreed — the client hid the RC host while the server still required it — producing a completion the server refused over a field no longer on screen.

**Imports.** `--dir-cache-time=0` imported as the seven-day default and `--links=false` imported as *on*. Zero is a value now, and an inline boolean is honoured.

Two findings were verified and deliberately **not** acted on, with the reasoning recorded in the review threads: rejecting a symlinked cache root would break the ordinary "cache on a fast volume" setup while the `vfs`/`vfsMeta` scoping already bounds the damage, and `CacheMaxSize` remains a soft, per-VFS quota — a hard guarantee needs a dedicated volume or admission control, not a different number.

## 👷 Verification already done

Exercised against a real installation on Unraid, and against a container with `/dev/fuse` passed in:

- The daemon starts under PUID with FUSE available, and a mount comes up at the configured path.
- 5,482 of 5,499 library symlinks resolve through it. The remaining 17 point at ids no longer in the WebDAV tree and were already dead.
- The mount appears in the host mount table as `fuse.rclone ... allow_other` and a separate container can list it.
- Roughly 8.7 GB has since streamed through it at about 20 MB/s.
- Shutdown logs `Released the built-in rclone mounts before shutdown.` and the mount is gone before the process exits.
- A restart cycle in a FUSE container mounts, releases on stop, and remounts — with no *connection refused* in the log, which the readiness wait is there to prevent.
- A deliberately stalled mount — mounted, then its daemon `kill -9`'d, leaving `ENOTCONN` — is cleared on the next daemon start, before the restart. A live mount on the same path is left alone and reported instead.
- The shipped rclone is the pinned version; the installer refuses a release below 1.70.3, an unsigned `SHA256SUMS`, and one signed by any other key.
- Backend suite 5,400 passed with 9 skipped; frontend 1,100 passed; rclone installer fixtures 15 passed; Release build with analyzers clean.

Two things are covered by tests rather than by eye, and are worth a look before merge: the always-expanded cache fields and the new wizard step have not been viewed in a browser.

The image was rebuilt and smoke-tested after the last round: rclone v1.75.1 and `fusermount3` present, backend reaching `Healthy`, frontend answering, and no error lines in the boot log.

## 🔒 Credentials

The WebDAV password cannot be read from configuration, which stores only a hash, so it is entered once and verified against that hash before any remote is created. It is stored only in rclone's own config, obscured, and never written to InfiniDysk's database or returned by any endpoint. The import reads mount settings only — it does not carry the password across, because nothing applies it yet and reading a credential that is never used only widens what a bug could leak. The setup wizard does not collect it either, which is why a wizard-configured mount comes up once the password is entered in Settings.

The daemon's own RC credentials are generated per process start, passed through the child process environment rather than the command line so nothing else in the container can read them from `ps`, and redacted from the log tail the admin UI shows.

## 🔗 Related links
- https://github.com/infinidysk/infinidysk/issues/254

https://claude.ai/code/session_01PRsdY7UxqNGTk1viNDL2EW

## Summary by CodeRabbit

* **New Features**
  * Added built-in rclone mounting for accessing the InfiniDysk WebDAV tree as local folders.
  * Added settings to configure, enable, disable, apply, remount, import, and monitor mounts.
  * Added WebDAV credential setup, migration previews, daemon logs, status details, cache controls, and actionable setup warnings.
  * Added setup support for choosing built-in rclone or an external sidecar.
* **Documentation**
  * Added configuration, Docker/FUSE requirements, migration, and troubleshooting guidance.
* **Bug Fixes**
  * Added validation to prevent conflicting or unsafe mount and cache paths.
  * Improved cleanup and protection for multiple configured mount locations.


samtheruby:
@coderabbitai full review

hoivikaj:
@samtheruby The body of the PR states "Its not ready" but my review was requested and PR was taken out of draft mode. Is this waiting on me for something?

samtheruby:
@hoivikaj it's not ready it's out of draft so that coderabbit could review it

samtheruby:
@coderabbitai full review

samtheruby:
@coderabbitai full review

hoivikaj:
@samtheruby this may be too late, have your agent review if any are still valid as you have been working- I asked my agent to do a review earlier but just got back from dinner to post it.

# PR #1362 Review and Action Plan

Review of [PR #1362](https://github.com/infinidysk/infinidysk/pull/1362), addressing [feature request #254](https://github.com/infinidysk/infinidysk/issues/254), on 2026-09-10.

Reviewed head: `e7c3d918b3f58c4c89c34855a4233593330daa31`. Base and merge-base: `699e0ca95ca9c196bd81314b3a8cf3cc2a0f159a`. Scope: 82 changed files, 11,689 additions, and 185 deletions. Source links below are pinned to the reviewed head so individual findings can be shared as GitHub comments.

## Findings

`P1` means a high-severity safety issue to fix before deployment or merge. `P2` means a functional or compatibility defect to resolve before advertising the affected workflow as supported. Executed probes are distinguished from static traces and checks that still require Linux/FUSE.

### R1 [P1]: Clear Cache Can Delete Databases And Unrelated Files

Sources: [RcloneCachePurger.cs](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneCachePurger.cs#L30), [cache-directory validation](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Config/ConfigManager.cs#L759), and [ConfigUpdateService.cs](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/ConfigUpdateService.cs#L20).

The cache directory can be any absolute path outside the mount list supplied in that request. `Purge` rejects a filesystem root but recursively deletes **every child** of any other directory. Setting the cache directory to `CONFIG_PATH`, a shared data directory, or an ancestor of other application data therefore turns Clear Cache into deletion of databases, backups, configuration, or unrelated files. The offline branch also permits deletion without a running daemon.

**Reproduced:** a temporary synthetic config directory passed validation; a non-cache database sentinel inside it was deleted; the purge returned no error. No real configuration or media was touched.

There is a second validation gap: `StageAsync` validates only incoming items. Saving a cache path inside an already configured mount succeeds when the request contains only the cache key. Supplying the same cache path and mount together is rejected. The claimed cross-setting protection therefore depends on request shape.

**Required change:** use a dedicated, identifiable cache namespace and delete only owned VFS data. Validate the merged effective configuration, including environment-owned values, and reject protected directories, unsafe ancestor relationships, and symlink aliases. Repeat ownership checks at deletion time. Tests must preserve databases, unrelated siblings, and files outside the owned cache under both online and offline clearing.

### R2 [P1]: The Installer Can Select A Checksum Outside The Signed Message

Source: [signature verification and checksum selection](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/scripts/install-rclone.sh#L141).

The installer verifies a clearsigned message but then uses `awk` to select the archive checksum from the **entire original file**. Successful signature verification does not authenticate arbitrary text appended after the signature.

**Reproduced with the real verifier:** take the authentic signed v1.75.0 checksum list and append an unsigned synthetic checksum entry named `rclone-v1.75.1-linux-amd64.zip`. GnuPG returns exit code 0 and the expected `VALIDSIG`; the installer's selection logic finds exactly one matching entry and selects the unsigned checksum. No archive was downloaded or executed in this probe.

Someone controlling the download responses could consequently substitute an archive and its matching unsigned checksum without forging the release key. This defeats the additional authenticity guarantee the installer is intended to provide, despite the pinned version and fingerprint.

**Required change:** extract the authenticated cleartext through GnuPG, require successful verification and the expected signer, and select the checksum only from that authenticated output. Do not rely solely on a `VALIDSIG` match in a pipeline. Add fixtures for unsigned prefixes/suffixes, a valid signed older release plus an appended requested filename, duplicate entries, and failed verification after status output.

### R3 [P1]: Stale-Mount Cleanup Can Detach A Live Mount

Source: [RcloneStaleMountCleaner.CanEnumerate](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneStaleMountCleaner.cs#L212).

The PR describes an `ENOTCONN`-specific safety check. The implementation instead converts **every** `IOException` and `UnauthorizedAccessException` into an unresponsive result, which causes the caller to attempt `fusermount3 -u -z` for a matching FUSE mount. A live sidecar with different ownership, a restrictive directory, or a transient I/O failure can meet those conditions without its daemon being dead.

**Reproduced:** a permission-denied temporary directory, represented as a FUSE entry in a synthetic mount table, produced `responsive=false` and one unmount attempt. No actual mount was changed. Whether the helper can detach a particular external mount depends on ownership and privileges, but the safety decision is already incorrect.

**Required change:** distinguish responsive, positively stale, and unknown/inaccessible states. Automatically detach only a positively identified dead transport, with the configured path and current mount identity still matching. Leave access denial, unrelated I/O errors, and timeouts alone. Add errno-specific tests and a real Linux test where another user can still read the live mount. Expected failures should produce contextual single-line warnings, not exception dumps.

### R4 [P2]: Save And Apply Do Not Apply Changed VFS Options

Sources: [the reconciler's keep decision](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneMountReconciler.cs#L84), [supervisor signature](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneDaemonService.cs#L602), and [status construction](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneMountsStatusBuilder.cs#L59).

An existing mount is retained whenever its mount point and remote match. Changes to cache mode, directory-cache time, cache age, read-ahead, links, access options, and size limits are not compared with the running mount. The supervisor notices some configuration changes but invokes this same no-op path; Apply does too. The install-wide cache-size limit is additionally absent from the supervisor's signature.

**Reproduced:** an in-memory RC client representing an existing root mount received zero mount calls and zero unmount calls after changed VFS settings were supplied; reconciliation returned zero errors. Status rows report configured options rather than proving those options are active, making the discrepancy difficult to detect.

**Required change:** compare normalized effective settings with an applied snapshot or actual VFS options. Remount only affected mounts, or explicitly report a pending change requiring a disruptive remount. Include global settings in the effective comparison. Test edits after an initial successful mount, including a global-limit-only edit; creation-time dictionary assertions do not cover this failure.

### R5 [P2]: A Failed Manual Remount Can Remain Down Indefinitely

Sources: [healthy-signature early return](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneDaemonService.cs#L543) and [manual remount transition](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Api/Controllers/RcloneMounts/RemountRcloneMountController.cs#L44).

After a successful supervisor pass, unchanged configuration suppresses all further mount reconciliation. The manual remount and cache-clear endpoints run their own reconciler but do not update the supervisor's success/failure state.

A concrete failure sequence is: normal mount succeeds; the user requests Remount; unmount succeeds; the replacement mount fails transiently. The endpoint reports the failure, but the supervisor still remembers a successful unchanged configuration and skips future passes. An individual mount removed while `rcd` stays alive is similarly not rediscovered. Watching the process is not sufficient to supervise its mounts.

**Required change:** route operations and their results through the supervisor, invalidate its applied state after destructive transitions, and reconcile actual mount state periodically with bounded backoff. Preflight remote availability before discarding a working mount, and define recovery if replacement fails. Test this exact sequence with a transient RC failure and verify recovery without another click, config edit, or daemon restart.

### R6 [P2]: Directory Invalidation Assumes One Daemon And One VFS

Sources: [VfsInvalidationTarget](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Database/DavDatabaseContext.cs#L1149) and [ForgetVfsPaths](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Clients/Rclone/RcloneClient.cs#L161). Contract: [rclone vfs/forget](https://rclone.org/rc/#vfs-forget).

Two newly relevant cases fail:

- With multiple distinct VFS instances, the request has no `fs` selector. Rclone requires that selector when more than one VFS is active. Retrying the same ambiguous request cannot fix it. Subtree mounts also require paths relative to their remote root, not blindly reused whole-tree paths.
- Merely starting the built-in daemon makes it the exclusive invalidation target, even before credentials or a working mount exist. The still-running external sidecar then stops receiving notifications. This is reachable during the documented enable/read/import workflow and affects continued use of both deployment modes.

These are static traces checked against rclone's documented contract. With a seven-day directory cache and polling disabled, failed invalidation can leave newly imported files invisible or deleted entries visible for a long time.

**Required change:** invalidate every supported active consumer, identify each VFS explicitly, and translate paths for its remote root. Preserve external notifications through cutover. Alternatively, restrict the initial release to one VFS and make exclusive-mode transitions explicit. Add real-protocol tests with root and subtree VFS instances, plus a sidecar still serving while built-in setup is incomplete.

### R7 [P2]: WebDAV Credential Changes Have No Usable Repair Path

Sources: [remote creation](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Api/Controllers/RcloneMounts/SetRcloneWebdavCredentialsController.cs#L62), [daemon startup settings](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneDaemonService.cs#L635), and [credential-form visibility](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/frontend/app/routes/settings/rclone/builtin-mount.tsx#L358).

The remote captures a WebDAV username, password, and URL once. Changes to those values do not recreate the remote or the mounted backend. The UI only offers credential entry when the remote name is absent, not when it exists but can no longer authenticate. Remote existence and mount-table presence can therefore look healthy while reads using the old credentials fail.

Even calling the credential endpoint directly is not a complete recovery design: it replaces configuration and invokes the reconciler, which retains existing mounts under R4 rather than ensuring the active backend has reloaded credentials.

**Required change:** detect authentication/configuration drift, expose an always-available reconnect action, and rebuild affected active backends under lifecycle coordination. Test username/password rotation and listener changes, including previously cached sessions. For the longer-term zero-duplicate-secret goal in #254, evaluate scoped internal service authentication; do not solve it by making the stored user-password hash reversible or exposing an unauthenticated listener.

### R8 [P2]: Re-running Setup Discards Existing Mount Definitions

Source: [completionSetupConfig](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/frontend/app/routes/setup/setup-model.ts#L212).

Every built-in symlink completion replaces the mount JSON with `builtinMountsFor(rclone.mount-dir)`, unless the entire mount key is environment-managed. It does this even when the saved configuration already has valid mounts and the user made no mount changes.

**Reproduced:** an unchanged two-mount configuration produces a one-mount completion payload. The additional mount will be removed by reconciliation, and custom settings on the retained root are replaced in persisted configuration. An unchanged wizard version only avoids automatic re-prompting; it does not protect an operator who intentionally revisits setup.

**Required change:** generate the initial root mount only for an empty configuration or an explicitly confirmed replacement. Preserve existing profiles and tuning on a no-op wizard run. Show a before/after review of additions, removals, and path changes. Test a multi-mount rerun, a tuned single mount, and an environment-managed mount list.

### R9 [P2]: Setup's Built-in And External Branches Disagree

Sources: [client-side branch validation](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/frontend/app/routes/setup/setup-model.ts#L229), [mode selection](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/frontend/app/routes/setup/setup-steps.tsx#L285), and [server-side branch validation](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/SetupWizardService.cs#L176).

The UI hides and stops validating external RC fields in built-in mode, but the server still requires an external host whenever `rclone.rc-enabled` is true. A user can select external mode, enable notifications without completing the host, then select built-in mode. The frontend accepts the playback step while submitting built-in enabled, external notifications enabled, and an empty external host; completion is rejected for a now-hidden setting.

**Reproduced at the client boundary:** step validation returned no errors for that state, and the completion payload retained the conflicting external RC value. The server rejection follows directly from `ValidateBranch`.

There is also no durable distinction between an explicit external choice and strategy-derived enablement: returning from STRM to symlinks unconditionally selects built-in mode when unmanaged.

**Required change:** model the selected deployment mode explicitly and validate the same effective branch on client and server. Preserve deliberate external choices, retain external credentials without letting irrelevant fields block built-in setup, and validate environment-controlled conflicts before Apply. Cover both mode-switch directions and an explicit external choice across a STRM round trip.

### R10 [P2]: Import Silently Changes Important Mount Semantics

Sources: [live import translation](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneImportTranslator.cs#L45), [boolean flag parsing](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneMountCommandParser.cs#L95), and [preview-to-form conversion](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/frontend/app/routes/settings/rclone/builtin-mount.tsx#L778).

The migration promise is stronger than the actual round trip:

- Live `DirCacheTime=0` and `CacheMaxAge=0` become seven-day defaults. The translator treats zero as absent; the frontend conversion also drops zero-duration values through truthiness checks.
- `--allow-other=false --links=false` imports as both values true, with no unsupported-flag warning. The parser recognizes the option names but ignores their explicit boolean values.
- Live options such as `ReadOnly`, chunk sizing, and cache-poll interval are not preserved or identified as unsupported. `ReadOnly` is already present in the RC response model but has no representation in the new mount configuration. The live import also assumes every reported mount should become an InfiniDysk mount, without establishing which source mounts actually belong to it.

The zero-duration and explicit-false cases were executed against the PR assembly. The omitted options and remote-identity assumptions are visible in the translation path.

**Required change:** distinguish omitted, zero, false, and unlimited values. Preserve supported policies exactly and explicitly block or disclose unsupported ones before cutover. Select the relevant source mount instead of assuming every remote is equivalent. A structured import is preferable to expanding the hand-written shell parser indefinitely; quoted arguments and overflow must at least fail cleanly. Do not add arbitrary shell execution as a workaround.

### R11 [P2]: The Symlink Coverage Warning Accepts Broken Mappings

Source: [coversSymlinkRoot](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/frontend/app/routes/settings/rclone/builtin-mount-model.ts#L136).

This function treats any ancestor mount point as covering `rclone.mount-dir` and ignores `RemotePath`. Filesystem containment is not equivalent to exposing the right virtual path.

For example, with `rclone.mount-dir=/mnt/library`, mounting the WebDAV root at `/mnt` puts `.ids` at `/mnt/.ids`, not `/mnt/library/.ids`. Likewise, mounting only `/content` at `/mnt/library` does not expose the root `.ids` tree there. Both cases currently suppress the warning.

**Reproduced:** the actual TypeScript function returned true for both invalid mappings.

**Required change:** validate the mapping from the expected `.ids` and completed-import paths into the mounted remote, including required link behavior. For a simple root-mount contract, require the root remote at the configured root path rather than treating all ancestors as equivalent. Test valid root mappings, ancestor mappings, subtree mounts, and missing symlink support; verify an existing imported file through its actual absolute target.

### R12 [P2]: Unchanged Settings Can Look Dirty Or Be Retuned On Blur

Sources: [mount/default comparison](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/frontend/app/routes/settings/rclone/builtin-mount-model.ts#L290) and [read-ahead input](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/frontend/app/routes/settings/rclone/builtin-mount.tsx#L516).

Two no-edit cases are broken:

- The wizard writes a minimal mount without `ReadAheadBytes`. The backend supplies its 512 MiB default, but the frontend signature turns the absent field into null. An unchanged saved wizard mount consequently differs from its server row, leaving Apply disabled with a misleading save-first warning.
- The input displays rounded bytes and compares the parsed display value with the original exact number on blur. For an imported value of 12,345,678 bytes, the displayed `12 MB` parses as 12,582,912. Merely focusing and leaving the field writes that different value. Comparing those two numbers does not establish that the user edited the text.

**Reproduced:** `mountsMatchServer` returned false for the unchanged wizard/default pair; the format/parse round trip produced the changed byte count above. The latter remains present despite the earlier resolved review thread about this behavior.

**Required change:** normalize absent defaults consistently while preserving explicit null/zero semantics, and track actual input edits independently of rounded display formatting. Test wizard-created mounts after save/reload, no-op focus/blur, imported non-round byte counts, and explicit read-ahead disablement.

### R13 [P2]: Cache Sizing Does Not Enforce The Advertised Database Reserve

Sources: [budget calculation](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneCacheBudget.cs#L70), [per-mount limit application](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneMountReconciler.cs#L241), and [operator guidance](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/docs/configuration/rclone.md#L70). Contract: [rclone VFS file caching](https://rclone.org/commands/rclone_mount/#vfs-file-caching).

The reserve calculation can produce an unsafe or negative budget, after which `Math.Clamp` raises it to 256 MiB. The direct probe returned a 268,435,456-byte allowance with zero free bytes. This contradicts the claim that the reserve cannot be consumed.

More fundamentally, `CacheMaxSize` is a soft, per-VFS quota: rclone cannot evict open files and checks quotas periodically. Sending the same install-wide value to independent mounts is not an aggregate cache-directory limit. Comparing mount-root strings also does not establish that separate Docker bind mounts have different physical backing storage.

**Required change:** do not enable a disk-backed cache below a safe reserve merely to meet a preferred minimum. Define behavior for low space, unknown space, and multiple caches; use real filesystem identity where needed. Consider a dedicated cache volume and rclone's minimum-free-space option, but document that it is also a soft target. A hard guarantee requires additional isolation or admission controls. Test sustained reads of files larger than the configured ceiling and several active profiles on the database volume.

## Recommendation

**Request changes. Keep the feature direction, but do not merge or deploy this snapshot as a generally safe built-in mount implementation.** R1-R3 are safety blockers. R4-R13 affect the workflows used to justify the feature and need fixes, explicit restrictions, or removal from the initial scope.

The author explicitly confirms in the PR discussion that the work is not ready and was taken out of draft for automated review. This is a pre-merge design and correctness review, not a suggestion that a human approval is the only remaining step.

## Design Assessment

### What Fits The Project

- Running the existing rclone engine is a good match for InfiniDysk's virtual-filesystem model. It avoids reimplementing FUSE/VFS behavior and removes deployment/configuration duplication for a common symlink-based installation.
- Opt-in backend enablement, separate external-client settings, a loopback RC listener, generated RC credentials, and no database migration are appropriate compatibility choices.
- Explicit mount reconciliation, refusing non-empty mount points, bounded restart attempts, cancellation-aware operations, and testable process/RC abstractions are useful foundations. Several earlier review problems, including failed-list handling and missing startup-option restarts, have been corrected.
- Disk-backed read-ahead and a zero per-transfer memory buffer can help concurrent playback. Warm cache reuse can reduce repeated Usenet reads. Those advantages depend on correct cache ownership, capacity handling, and measured behavior under load.

### What Should Change In The Approach

**Recommended first release: optional, Linux/Docker-focused built-in mode with one verified root mount, one owned cache namespace, and an explicit external alternative.** Keep a future-compatible mount-list schema, but do not equate that schema with proven support for independent live profiles. Multi-mount invalidation, cache isolation, and option application must be demonstrated before advertising that use case.

The current PR combines process supervision, stale-kernel-mount cleanup, arbitrary mount profiles, sidecar migration, credential entry, a new setup default, and destructive cache maintenance. These capabilities multiply one another's failure modes. The runtime ownership and safety model should be stabilized before expanding all of the UI workflows at once. This does not require discarding the implementation or introducing a large new framework.

There should be **one lifecycle owner** for process replacement, mount changes, credential replacement, and cache maintenance. The semaphore currently covers reconciliation calls but not every process transition. In particular, [StopAsync](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/backend/Services/RcloneDaemonService.cs#L154) unmounts before cancelling the supervisor and outside its mount gate; settings-driven process teardown also bypasses that gate. This leaves interleavings that need barrier-controlled tests, even though a leftover kernel mount from those races was not reproduced locally.

A conservative design would use a validated effective-config snapshot, a daemon generation identifier, serialized commands, explicit desired/applied/observed mount states, and bounded reconciliation of drift. A failed replacement should have a defined recovery path. Avoid holding the process-state lock across slow shutdown I/O or allowing a queued operation to act on an obsolete daemon generation.

**Do not make an unverified built-in mount the unconditional symlink setup default.** The wizard currently completes without collecting the WebDAV password or checking mount readiness; a user must later discover Settings and finish connecting. It also cannot make missing FUSE privileges or host mount propagation disappear. Either finish and verify the built-in path inside setup, or label it incomplete and require an explicit choice with a usable external fallback.

The original issue also asks to eliminate duplicated WebDAV credentials. Re-entering the same password into a separate rclone config, without rotation handling, does not finish that requirement. A one-time-entry MVP can be acceptable if the scope change is explicit and reconnect works; scoped internal authentication is a separate design decision that deserves security review.

### Use-Case Impact

| Use case | Improvement | Risk or restriction to address |
| --- | --- | --- |
| New Linux Docker installation using Sonarr/Radarr symlinks | Fewer services and less manual rclone configuration | Credentials, FUSE access, target paths, permissions, and propagation must be verified before setup is presented as complete. |
| Existing external sidecar with built-in mode off | Existing configuration is retained | Preserve this behavior and add an explicit regression test; do not silently change mode through setup. |
| Gradual migration while the sidecar still serves users | Read-only preview and path preservation are useful | Starting the new daemon steals invalidations; stale detection can target a live mount; migration can retune policies. |
| Plex playback and Arr analysis using different mount profiles | Potentially useful independent cache policies | Multiple VFS targets, effective option application, cache independence, and aggregate disk use are not yet established. |
| Large media, simultaneous playback, or constrained NAS storage | Warm reads can save bandwidth; disk read-ahead can reduce memory pressure | Cache quotas are soft, open files can exceed them, and scans/prefetch can consume additional bandwidth and disk I/O. |
| Rootless or security-restricted containers, native archives, remote media hosts | External mode remains a useful escape path | Built-in mounting requires additional deployment capabilities; a local mount does not automatically become visible on another host or in another container. |
| STRM or direct WebDAV clients | No rclone is required for these paths | Keep them independent of FUSE privileges, disk caching, and mount setup. Scope the documentation's filesystem prerequisite accordingly. |
| Maintenance, upgrades, and daemon restarts during playback | Centralized operations are easier to discover | Mounts now share the app's lifecycle. Unmounting can invalidate open handles; do not promise that every player will merely pause and transparently resume. |

Keeping the external option is not just backward compatibility. It preserves privilege separation and independent mount lifecycle for operators who deliberately chose them. Adding FUSE privileges and relaxed AppArmor to the combined application container has a security cost that should be explicit.

## Action Plan

### Phase 1: Close The Safety Boundaries

Address R1-R3 before additional deployment testing. Disable Clear Cache until it can prove ownership of every deletion target. Validate complete effective settings, not only changed keys. Extract only authenticated checksum text. Restrict stale cleanup to positive dead-transport evidence.

Acceptance: unrelated synthetic data survives all purge cases; unsigned checksum material is rejected even beside a valid signature; an inaccessible but live FUSE mount is never detached. Include negative cases, not just successful clearing or an abstract `responsive=false` fake.

### Phase 2: Make Runtime State Authoritative

Address R4-R7. Put start/stop/reconfigure/remount/reconnect/cache maintenance under one lifecycle owner. Track effective settings and daemon generations, reconcile disappeared mounts, and feed endpoint failures into bounded retry state. Check prerequisites before destructive replacement. Implement per-VFS invalidation or explicitly constrain the initial release to one VFS, preserving external notifications during transition.

Acceptance: changing every exposed setting either changes the running behavior or reports a clear pending operation; a transient failed remount recovers automatically; old-generation requests cannot mutate a replacement daemon; credential rotation has a working repair flow. Test shutdown, cache clear, and settings changes while another mount operation is blocked at a known barrier.

### Phase 3: Preserve User Intent And Migration Data

Address R8-R12. Preserve existing profiles on wizard reruns, model the selected rclone mode independently of import strategy, and align client/server validation. Preview any mount removals or policy changes. Preserve explicit zero/false values and exact byte counts. Validate the actual symlink mapping.

Apply the repository's setup-impact requirements to the new keys: pin the entire environment-managed mount configuration in the UI, validate it together with writable settings, and reject conflicts with the organized library directory as well as the symlink root. The current server branch checks only the configured symlink mount directory, not every proposed built-in mount. Decide whether setup remains incomplete until authentication and a real mount check succeed; do not bump the wizard version merely to prompt existing users into an unfinished optional path.

Acceptance: an unchanged existing installation produces no configuration changes; explicit external choices survive navigation; hidden external settings cannot block built-in completion; managed values are visibly locked; the imported configuration either preserves behavior or clearly explains and requires acceptance of each difference.

### Phase 4: Make Capacity And Operator Status Honest

Address R13 and the operational gaps exposed by the review. Define low-space behavior and whether the UI's limit is per mount or aggregate. Report effective limits and observed mount/authentication failures rather than only saved values or PID existence. Warn that remount/cache clearing may stop playback, and keep destructive operations explicit.

Finish these adjacent API/UI checks:

- Handle non-success HTTP responses and the standard `error` response in every action. [Apply currently parses any JSON response as a successful operation-shaped result](https://github.com/infinidysk/infinidysk/blob/e7c3d918b3f58c4c89c34855a4233593330daa31/frontend/app/routes/settings/rclone/builtin-mount.tsx#L146); a JSON 4xx/5xx without an `errors` array can render a success message. Remount and credential submission need the same scrutiny.
- Exercise the status controller through its actual eight-second deadline. The transport rethrows caller cancellation; merely proving the deadline token cancels does not prove the endpoint returns the documented partial status.
- Keep expected connection, permission, and shutdown-timeout failures as actionable, throttled log events. Preserve stacks for unexpected faults.
- Update the documentation for the new install-wide cache-size control, effective override precedence, the now-editable directory-cache duration, credential recovery, and supported deployment modes. The current guide still says there is no separate limit key and that directory-cache time is not exposed in the UI.

Acceptance: operator-visible success means the requested action succeeded; pending, failed, unauthorized, low-space, and interrupted states remain distinguishable. Review the setup and cache controls in a browser on desktop and mobile, including managed settings and long paths.

### Phase 5: Prove The Supported Deployment And Pass The Gates

Use the existing xUnit, Vitest, and installer-fixture conventions. Add focused regressions for the findings rather than relying on another broad happy-path count. Then exercise the actual packaged rclone version in Linux/FUSE with synthetic WebDAV content and no live Usenet credentials.

| Required integration check | Acceptance condition |
| --- | --- |
| Fresh container with supported PUID/PGID and FUSE settings | Mount authenticates, expected virtual paths resolve, and a separate consumer container can read an existing imported symlink target. |
| Missing privileges, missing target directory, or incorrect propagation | Setup gives a specific actionable state; it does not claim a ready shared filesystem. |
| Multiple supported profiles, or the enforced single-profile restriction | No ambiguous VFS invalidations, unintended shared cache state, or silent option differences. |
| Normal shutdown, hard kill, and restart | Owned dead mounts recover; live/inaccessible foreign mounts are untouched; process teardown does not race remount. |
| Cache clear/remount/settings edit during active streaming | Documented interruption behavior, protected persistent data, and deterministic recovery without a second operator action. |
| Credential rotation and sidecar cutover/rollback | New reads authenticate, old paths remain valid, and the correct consumers continue receiving invalidations. |
| Large open files and concurrent reads near the disk reserve | Capacity behavior matches the documented policy and cannot be mistaken for a hard rclone quota guarantee. |

Resolve the current coverage-ratchet failure by raising the measured floors, not weakening tests or lowering the gate. For the reviewed run, it asks for branches `27 -> 30`, lines `31 -> 34`, and statements `30 -> 33`. Re-evaluate the required values after the fixes because new tests can move coverage again.

Complete the introducing-release documentation and scope statement, obtain human review, and leave merging to a maintainer. The original feature issue should only be closed when the accepted credential, setup, external-mode, and mount-profile scope is clear.

## Verification

### Executed During This Review

The PR was inspected in a separate detached worktree. Backend probes used the assembly built from that snapshot; frontend probes imported the actual TypeScript model functions. Filesystem probes used newly created temporary fixtures that were removed afterward. The signature probe used public release material and a temporary public-key keyring.

| Probe | Observed result |
| --- | --- |
| Synthetic config directory used as cache | Validation accepted it; a non-cache database sentinel was deleted; purge error was null. |
| Cache nested beneath a mount | A cache-only update passed; the equivalent combined cache/mount request was rejected. |
| Signed older checksum list plus unsigned requested entry | GnuPG exited 0; expected signature accepted; exactly one unsigned synthetic checksum selected. |
| Permission-denied directory represented as a live FUSE entry | Marked unresponsive; one unmount attempted against the fake helper. |
| Existing mount with changed VFS settings | Zero mount calls, zero unmount calls, zero reconciliation errors. |
| Explicit zero imported durations | Both became 604800 seconds. |
| Explicit false imported boolean flags | Both became true; no unsupported-flag warning. |
| Invalid ancestor and content-only symlink mappings | Both reported as covering the configured root. |
| Built-in setup with unfinished external RC settings | Playback validation passed; completion retained built-in enabled and external RC enabled with an empty host. |
| Unchanged two-mount setup rerun | Completion replaced two mounts with one. |
| Unchanged wizard/default settings comparison | Returned false, causing a spurious unsaved-state indication. |
| Unchanged read-ahead display round trip | 12345678 bytes became 12582912 bytes. |
| Automatic cache budget with zero free bytes | Returned 268435456 bytes. |

These probes confirm local logic and data-contract failures. They are not claims that real FUSE detachment, media-server playback, or cross-container propagation was executed locally.

### CI And Local Test Limits

The targeted local xUnit attempt selected `RcloneMountReconcilerTests`, `RcloneCachePurgerTests`, and `RcloneBuiltinConfigTests`. It did **not** reach test execution. The normal build encountered analyzer errors in unchanged RapidYencSharp code, including CA1707 and CA5393. A diagnostic rerun with analyzers disabled then encountered CS0012/CS0534 library-reference errors in the test project. Those unrelated build problems were not repaired or presented as PR findings.

At the reviewed head, [CI run 34526182132](https://github.com/infinidysk/infinidysk/actions/runs/34526182132) reports backend tests, backend quality/formatting, HTTP contracts, library/native lanes, Docker smoke build, and quality-ratchet checks passing. CodeQL also reports success. The frontend job and required quality gate fail because the **coverage floors need to be raised**, not because coverage regressed:

| Metric | Actual | Stored floor | Requested floor |
| --- | --- | --- | --- |
| Branches | 32.23% | 27% | 30% |
| Lines | 36.59% | 31% | 34% |
| Statements | 35.31% | 30% | 33% |

Failure log: [frontend coverage-ratchet job](https://github.com/infinidysk/infinidysk/actions/runs/34526182132/job/103035526527). CI success on the other lanes does not exercise the missing state-transition and destructive-boundary cases above. The author's reported Unraid/FUSE and streaming checks were not independently repeated, and no browser visual verification was performed during this review.

### Review Context And Scope

The PR description, original feature request, human discussion, and all 69 available review threads were inspected. Current code was checked rather than treating every old comment as still valid. For example, the earlier disposed-client exception claim is not supported by the present `RcloneClient.Dispose`, which only removes its configuration subscription and does not dispose the shared transport. Conversely, the rounded read-ahead blur problem remains reproducible despite being marked resolved.

The deliverable is this review and remediation plan. Application source in the user's checkout was not changed, and no review comments were posted or pull requests merged.

samtheruby:
@coderabbitai full review

samtheruby:
@coderabbitai full review

sam-rubinstein:
PR should be ready @hoivikaj

samtheruby:
PR should be ready @hoivikaj

hoivikaj:
Something like this would roll in a new feature version, going to quickly workthrough some bug fixes for teh 1.4 series first. Hang tight. I don't mind doing the housekeeping on the PR as it drifts until then. @samtheruby

samtheruby:
sounds good let me know if you want me to change anything

hoivikaj:
Review sweep completed in c884ab14.

Fixed: unsupported external VFS cache modes now import as `Off`; unrepresentable unlimited cache-age/read-ahead values are surfaced in the preview; and new mount IDs no longer collide after a removal.

Resolved as already covered: import flag warnings, request-abort-safe reconciliation, cache budgeting/free-space behavior, cache-root symlink handling, and hosted-service ordering.

Dismissed after verification: loopback-only backend WebDAV HTTP is intentional; `v1.4.0` is published; the fenced documentation blocks are valid; the semicolon-only C# collection declaration is valid; and the reported disposed-client race does not dispose the shared transport.

Focused validation passed: 31 backend rclone-import tests and 45 frontend mount-model tests. No GitHub Code Quality or GitHub Copilot review threads were present.

samtheruby:
Might've found a bug that needs fixing but I probably wont get around to it until this Sunday/Monday, don't merge this until then, or if you want to fix it the bug is that there is no usermod ensuring that everything in rclone is owned by the program's user, causing certain tasks to fail, such as removing orphan files.

hoivikaj:
@samtheruby do you have any details on the bug you encountered?

I am leaning towards really looking into this when my credits reset in the new month, so you have some time to tinker.

PR #1371, opened 2026-09-09
fix(repair): stop a busy PAR2 repair from freezing every health-check worker

samtheruby:
## Problem

Only one PAR2 repair runs at a time (`Par2RepairService._repairAdmission`, a `SemaphoreSlim(1, 1)`). A health-check worker that found damage awaited that semaphore with no timeout, and it kept its `_inProgress` slot for the whole wait.

One release whose articles are all gone could therefore park every worker. Observed on a live instance: a repair started at 21:34:50 UTC and was still running 15 minutes later with `BytesRead` 0, grinding through ~97 failed BODY fetches per minute, while metrics showed

```
nzbdav_par2_repair_active 1
nzbdav_par2_repair_admission_waiters 2
nzbdav_health_check_gate_operations{state="active"} 0
```

Three of four workers were sitting in PAR2, no STAT work was in flight, and the last completed health check was three seconds after that repair began. With `repair.healthcheck-workers` set to 4, no library file was checked for the whole window — the setting had stopped meaning "files checked at once", which is what the docs promise.

## Change

Inline callers (health-check workers and playback reports) now wait a bounded `Par2RepairService.AdmissionWaitTimeout` (5s) and get the new `Par2RepairOutcome.Deferred` if the slot does not free up. The background repair queue keeps waiting indefinitely, so queued jobs are never silently dropped.

`Deferred` is deliberately not folded into `NotRepaired`. A busy repair slot says nothing about whether parity can fix the file, and `NotRepaired` falls through to the *Arr replacement path — which would remove and re-grab a download that parity might still recover. All three health-check call sites now defer the item instead: `HealthRepairPending` is set, the next attempt is scheduled 15 minutes out (preserving the `UnixEpoch` urgent sentinel), and an `ActionNeeded` result explains why.

## Tests

- `Par2RepairAdmissionTests.InlineCaller_DefersWhenAnotherRepairHoldsTheSlot` — an inline caller returns `Deferred` instead of blocking. Fails with `TimeoutException` without the bounded wait.
- `Par2RepairAdmissionTests.InlineCaller_StopsWaitingOnceTheSlotIsFree` — a deferral gives up cleanly and never leaks a permit.
- `HealthCheckDegradedClassificationTests.DeferredPar2_HoldsItemForRetryInsteadOfReplacement` — a deferred repair marks the item pending and leaves the blob alone rather than entering the replacement path. Fails without the call-site guards.

Full suite on this branch: 5189 passed, 0 failed, 9 skipped (`dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release`). `dotnet format whitespace --verify-no-changes` reports nothing in the five changed files.

## Not in scope

Two related findings from the same investigation, left for separate PRs:

1. The PAR2 phase of a health check has no wall-clock deadline of its own — only the STAT sweep has one (`HealthCheckProgressTimeout`). The repair *owner* can still grind for a long time on a dead release; this PR only stops it from taking the other workers with it.
2. Discovery has no early-out for a release whose articles are all missing. It walks the recovery set until a byte or request budget trips, at roughly 600 ms per dead article.

Also present on a clean `main` checkout, unrelated to this change: `ArchitectureBoundaryTests.NonApiCodeDoesNotDependOnControllers` fails.

https://claude.ai/code/session_017B8WaHYfyRiuLJvBLYA9s3

## Summary by CodeRabbit

* **Bug Fixes**
  * Health checks now defer PAR2 repairs when another repair is active, rather than incorrectly sending files to replacement.
  * Deferred files remain queued, retain their original data, and are automatically retried after 15 minutes.
  * Health-check workers are released promptly when repair capacity is unavailable.

* **Documentation**
  * Updated repair guidance to explain deferred repairs and queue behavior.

### DIFF
--- a/backend/Services/HealthCheckService.cs
+++ b/backend/Services/HealthCheckService.cs
@@ -45,6 +45,12 @@ public class HealthCheckService : BackgroundService, IHealthCheckQuiescence
     private static readonly TimeSpan MissingPayloadConfirmedRecheck = TimeSpan.FromDays(7);
     private static readonly TimeSpan HealthCheckProgressTimeout = TimeSpan.FromMinutes(5);
 
+    /// <summary>
+    /// How long an item waits before another repair attempt after PAR2 reported that a
+    /// different repair owned the single-repair slot.
+    /// </summary>
+    internal static readonly TimeSpan Par2AdmissionDeferral = TimeSpan.FromMinutes(15);
+
     // Repeated remove-and-blocklist repairs for the same library path in a short window indicate
     // a replacement loop (Arr keeps re-grabbing a release repair keeps rejecting, issue #732).
     // After the limit is hit, further repairs at that path are deferred instead of deleting again.
@@ -1279,6 +1285,13 @@ public class HealthCheckService : BackgroundService, IHealthCheckQuiescence
                 await HandleUnreadablePayloadAsync(davItem, dbClient, exception, ct).ConfigureAwait(false);
                 return;
             }
+
+            if (par2Outcome is Par2RepairOutcome.Deferred)
+            {
+                await DeferRepairUntilOwnerAvailable(davItem, dbClient, ct).ConfigureAwait(false);
+                return;
+            }
+
             if (par2Outcome is not Par2RepairOutcome.NotRepaired)
             {
                 var utcNow = DateTimeOffset.UtcNow;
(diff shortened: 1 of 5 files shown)


hoivikaj:
@samtheruby I have an upcoming PR that addresses only this issue that isnt based on top of your rclone work (so wouldnt contain 88 file changes :D )

samtheruby:
Closed in favor of https://github.com/infinidysk/infinidysk/pull/1383

PR #1383, opened 2026-09-10
fix(health): defer contended PAR2 repairs

hoivikaj:
## Summary

This PR replaces #1371 with an independent Phase 1 implementation.

- Adds explicit `DeferredBusy` PAR2 outcomes.
- Makes inline health/playback repair admission non-blocking while preserving queued repair waiting and cleanup ownership.
- Prevents same-file and cross-file inline callers from waiting behind an active repair.
- Records deferred repairs as unhealthy/action-needed and preserves urgent retry intent.
- Suppresses repair candidates while PAR2 admission is busy without suppressing ordinary health checks.
- Updates contention tests for deferred behavior and stale-flight self-healing.

Phase 2 request-reduction and repair-discovery performance work is intentionally left separate for reviewability.

## Validation

- `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --no-restore --filter FullyQualifiedName~Par2RepairServiceMultipartTests`
- 47 passed
- Workspace diagnostics report no errors in the touched C# files

The terminal stopped returning output during a later combined validation command, so the focused 47-test run and source diagnostics are the reliable validation recorded here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved repair availability checks so new repair candidates are not selected when PAR2 repair capacity is unavailable.
  - Busy repair operations are now deferred immediately instead of blocking or being reported as failed.
  - Deferred repairs are marked pending, recorded as unhealthy where appropriate, and rescheduled shortly when non-urgent.
  - Missing-segment repair results now distinguish deferred work from successful repair or clean verification.
- **Tests**
  - Updated repair coverage for immediate deferral and retry behavior during concurrent requests.

### DIFF
--- a/backend/Services/HealthCheckService.cs
+++ b/backend/Services/HealthCheckService.cs
@@ -808,7 +808,7 @@ public class HealthCheckService : BackgroundService, IHealthCheckQuiescence
         if (davItem.NextHealthCheck != DateTimeOffset.UnixEpoch)
             davItem.NextHealthCheck = now.AddSeconds(1);
         Log.Warning(
-            "PAR2 repair deferred for {Path}; admission is busy ({Outcome}).",
+            "PAR2 repair deferred for {Path} because repair capacity is contended ({Outcome}).",
             davItem.Path,
             outcome);
         await RecordHealthResult(
@@ -816,7 +816,7 @@ public class HealthCheckService : BackgroundService, IHealthCheckQuiescence
                 davItem,
                 HealthCheckResult.HealthResult.Unhealthy,
                 HealthCheckResult.RepairAction.ActionNeeded,
-                "PAR2 repair remains pending because another repair is active.",
+                "PAR2 repair remains pending because repair capacity is contended.",
                 ct)
             .ConfigureAwait(false);
     }
--- a/backend/Services/HealthCheckService.cs
+++ b/backend/Services/HealthCheckService.cs
@@ -1459,6 +1459,11 @@ public class HealthCheckService : BackgroundService, IHealthCheckQuiescence
             ? await _par2RepairService.TryPar2RepairAsync(
                 davItem, holeSegmentIds, ct).ConfigureAwait(false)
             : Par2RepairOutcome.NotRepaired;
+        if (par2Outcome == Par2RepairOutcome.DeferredBusy)
+        {
+            await DeferPar2RepairAsync(davItem, dbClient, par2Outcome, ct).ConfigureAwait(false);
+            return;
+        }
         if (par2Outcome is Par2RepairOutcome.Repaired or Par2RepairOutcome.VerifiedClean)
         {
             var utcNow = DateTimeOffset.UtcNow;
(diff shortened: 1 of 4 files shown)


samtheruby:
@hoivikaj your fix is better then mine, so I'll close mine out. Just a quick comment from my claude review:

Design here is good — the `CanAcceptInlineRepair` gate in `RefillWorkerSlotsAsync` and `WaitAsync(0)` are better than a plain admission timeout, and tightening `is not NotRepaired` to `is Repaired or VerifiedClean` is the right shape. Four things before it lands.

**1. `DeferredBusy` is unguarded at one of the three call sites.** `HandleConfirmedHolesAsync` (~1459) only got the success test tightened, with no deferral branch, so a deferred repair falls through to:

```csharp
if (FilenameUtil.IsImportantFileType(davItem.Name))
    AddMissingSegmentIds(holeSegmentIds);
await Repair(davItem, dbClient, ct).ConfigureAwait(false);
```

That's remove-and-blocklist through *Arr plus seeding the fail-fast reimport cache, on a file PAR2 never looked at because the slot was busy — the exact case this PR exists to prevent. `HealthRepairPending` isn't set there either, so the item never comes back as a repair candidate. Needs the same `DeferPar2RepairAsync` branch as the other two sites.

**2. `Backend tests` is red.** Reproduced locally on the same base with `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release` — 5185 passed, 1 failed:

```text
Par2RepairServiceCorruptSourceTests.BackgroundRepair_UrgentCallJoinsFlight_AndCanceledJoinerDoesNotCancelOwner
Assert.ThrowsAny() Failure: No exception was thrown
Expected: typeof(System.OperationCanceledException)
```

This one is the test, not the code. It encodes the old contract where a joiner waits and can be cancelled mid-wait. With `if (!flight.Task.IsCompleted) return Par2RepairOutcome.DeferredBusy;` there's no `await` before the return, so the joiner completes synchronously and the test's `Cancel()` lands too late. (Adding `ct.ThrowIfCancellationRequested()` before the return doesn't fix it — I tried.)

This version passes and takes the full suite green (5186 passed, 0 failed):

```csharp
// An already-cancelled caller still observes its token.
using var canceledJoinerCts = new CancellationTokenSource();
await canceledJoinerCts.CancelAsync();
await Assert.ThrowsAnyAsync<OperationCanceledException>(
    () => release.Service.TryPar2RepairAsync(
        release.Item,
        [release.ContentSegmentIds[0]],
        canceledJoinerCts.Token));

// An inline caller never waits behind the running owner.
Assert.Equal(
    Par2RepairOutcome.DeferredBusy,
    await release.Service.TryPar2RepairAsync(
        release.Item,
        [release.ContentSegmentIds[0]],
        CancellationToken.None));

// Neither joiner disturbed the owner: it still finishes on its own.
allowSourceRead.TrySetResult(true);
using var ownerTimeout = new CancellationTokenSource(TimeSpan.FromSeconds(10));
while (release.Service.GetDiagnosticSnapshot().TotalSucceeded == 0)
    await Task.Delay(25, ownerTimeout.Token);
Assert.True(release.Store.Contains(release.ContentSegmentIds[0]));
Assert.Equal(1, release.Service.GetDiagnosticSnapshot().TotalSucceeded);
```

Worth renaming to something like `BackgroundRepair_InlineJoinersDeferWithoutDisturbingOwner`.

**3. Undeclared behaviour change.** `LateDifferentVolumeRequest_RunsFollowUpBeforeReturningSuccess` now asserts the follow-up does *not* run (1 succeeded job instead of 2, patch not created), while the name still says it does. A late request for a different volume of the same release used to be repaired within the same call. Probably a fine trade given the retry path, but worth stating in the description and renaming the test.

**4. Two small ones.** The third call site has the same `DeferredBusy` block twice (2996 and 3001 — the second is unreachable), and `Interlocked.Exchange(ref _admissionActive, 1)` is written inside the `QueuedWait` branch and again after the if/else.

PR #1406, opened 2026-09-13
fix(auth): sign in works again behind a reverse proxy

hoivikaj:
## Summary

Logging in through a reverse proxy failed with "Something went wrong while signing in." Root cause: React Router 8.3's built-in action CSRF check compares the SSR request URL's host against the browser's `Origin` header. `@react-router/express` falls back to the *raw* internal `Host` header's port whenever `X-Forwarded-Host` omits one (the common case for a default-port public origin, e.g. `https://example.com`). Since the frontend container listens internally on a non-default port (e.g. `:3000`), that port leaked into the reconstructed request URL, so the host never matched the plain `Origin` header and every proxied login was rejected.

The previous fix (#1401) only added throttled logging around the rejection — it never corrected the origin reconstruction, so proxied logins kept failing identically.

## Fix

- `normalizeForwardedHost` (`frontend/server/forwarded-headers.ts`) explicitly reconstructs the public origin from trusted `X-Forwarded-Proto` / `X-Forwarded-Host` / `X-Forwarded-Port` headers (in that trust order), rewriting the `Host` header before Express or React Router see the request — never falling back to the container's own port. Default ports (80/443) are omitted to match browser `Origin` canonicalization. Handles comma-separated multi-hop headers and bracketed IPv6 hosts.
- Wired into `frontend/server/app.ts` as middleware, gated by the existing `TRUST_PROXY` opt-in, running before the React Router request handler.
- This does not weaken CSRF protection — it only ensures the reconstructed origin the framework check compares against reflects the real public origin instead of a bogus internal one.

## Test plan

- Added unit coverage in `frontend/server/forwarded-headers.test.ts` for scheme/host/port reconstruction, default-port omission, multi-hop headers, IPv6, and no-op cases when `TRUST_PROXY` is disabled or headers are absent.
- Added `frontend/server/csrf-origin-regression.test.ts`: an end-to-end regression against a real Express server reproducing a reverse-proxy configuration (internal port + default-port HTTPS proxy, and a non-default forwarded port variant). Confirms both direct and proxied logins now succeed (200), while a genuine cross-origin request is still rejected (400), proving CSRF protection is preserved.
- `npx tsc -b` and `npx eslint --max-warnings 0` clean on changed files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of forwarded host, protocol, and port information when the application runs behind a proxy.
  - Fixed server-rendered requests and form actions that could incorrectly use an internal port instead of the public address.
  - Preserved protection against genuine cross-origin actions, which continue to be rejected.

- **Tests**
  - Added coverage for proxied requests, default and non-default ports, protocol handling, multi-hop forwarding, and IPv4 and IPv6 host formats.


hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai full review

hoivikaj (commented):
Fetching PR review state for #1406 to check whether there are requested changes.

hoivikaj (commented):
Requesting review metadata for PR #1406 to inspect whether there are changes requested and what comments are pending.

hoivikaj (commented):
Checking the current state of PR #1406 and any pending review actions.

PR #1407, opened 2026-09-13
fix(health): preserve urgent repairs across concurrent checks

### LINKED ISSUE #1388: Files whose first article has expired are invisible to repair
### Before submitting

### InfiniDysk version

1.3.0

### Installation method

Official Docker image

### Affected area

Queue, NZB processing, or health checks

### What happened?

Forgive me in advance. I had Claude help me diagnose the issue and write this up...

Running v1.3.0.

#1014 fixed dead-tail seeks by calling ScheduleRepair when a seek failure is caused by a short segment. There is a sibling case it doesn't reach: when the first article of a file is gone entirely. Those files are permanently invisible to repair, the failure counter, par2 and the Arr cycle — no matter how often they are read.

I have 125 files confirmed by direct probing, and 139 distinct files appeared in a single two-hour window once a Plex media-analysis sweep began touching them.

Signature. Byte 0 through roughly 768 KB — exactly one article — returns 404. Everything from 768 KB onward reads perfectly. The same boundary on a 254 MB file and on a 4 GB file. Ranged GETs at 10/25/50/75/90/99% all return 206.

Why the #1014 fix doesn't fire. In ExceptionMiddleware.cs:

if (e.InnerException is EndOfStreamException &&
    context.Items["DavItem"] is DavItem davItem)
{
    ScheduleRepair(davItem);
}

A short segment throws EndOfStreamException. A completely absent first article does not — the interpolation search cannot locate the position at all, and the inner reason is "Corrupt file. Cannot find byte position N". Different inner exception, so ScheduleRepair is never called.

Two hours of logs, with Repair After Streaming Failures = 0 (immediate):

217  "could not seek to byte position …"   (207 of them at position 0)
139  distinct files
  0  "Scheduled dynamic repair"
  0  "Deferring dynamic repair"

A second, separate observation. These files hold a current healthy verdict. 123 of my 125 have a most-recent HealthCheckResult of plain "File is healthy." — no sampling caveat — dated between 2026-09-06 and 2026-09-11, with NextHealthCheck set years into the future. A file whose first article answers 430 No Such Article is passing a health check that reports no limitation.

Net effect. This is the only damage class that accumulates. Damage after the first article gets found and replaced — I traced 395 par2 jobs, each followed by a health verdict and an Arr remove/blocklist within seconds. Damage in the first article is invisible to both. In #461's vocabulary this is the most fatal case there is — it is where the MKV EBML header and the MP4 ftyp/faststart moov live — and it currently receives no repair at all.

### Steps to reproduce

1. Find a library item whose first Usenet article has expired. (In my case they surfaced when Plex's media-analysis sweep began reading every file from offset 0; "could not seek to byte position 0. Reason: Corrupt file" in the InfiniDysk log identifies them.)
2. Confirm the shape — head dead, body intact:
curl -u user:pass -r 0-65535       http://infinidysk:3000/content/…/file.mkv   ->  404
curl -u user:pass -r 786432-851967 http://infinidysk:3000/content/…/file.mkv   ->  206
3. Note the 404 returns in about 1 ms, with Attempts: 0; CachedSkips: 5; BytesFetched: 0 in the segment log — no provider is contacted.
4. Repeat step 2 as many times as you like. I read these ~175 times across 45 minutes.
5. Check the outcome:
no "Scheduled dynamic repair" in the log
no row in Par2RepairJobs for the item
HealthCheckResults newest verdict still "File is healthy.", DavItems.NextHealthCheck still years out
Radarr/Sonarr never asked to replace it

Settings during the above: Repair After Streaming Failures = 0, PAR2 background repair enabled, Arr instances configured and working (the same install performed 1,107 successful health-check remediations).

### Expected behavior

A read that fails because the article covering the requested range is confirmed absent from every provider should enter the repair pipeline exactly as a short segment does — count a streaming failure, set the item urgent, run par2, and fall through to the Arr remove-and-blocklist when par2 cannot help.

Respecting the caution in #1014. That gate was deliberately narrowed because "other seek failures can be transient header-probe errors … and must not trigger Arr removal or blocklisting". That concern is real — in the same two-hour window, alongside 207 permanent "Corrupt file" seeks there were 31 circuit-open and 26 connection-limit reasons, which must never be burned on.

But the two are distinguishable by cause rather than exception type. A seek failure whose underlying cause is a confirmed article miss — a 430 across all eligible providers, or a negative-cache hit for that segment — is integrity-proven in exactly the way a short segment is. A connection failure or an open circuit is not. Gating on that would cover this case without reintroducing the transient-failure risk.

Secondarily, a health check that returns "File is healthy." for a file whose first article is absent should not do so, and should not then schedule the next check years out.

_No response_

### Environment

_No response_

### Additional context

_No response_

hoivikaj:
## Summary
- add revision-aware failure snapshots and conditional clearing
- serialize repair scheduling and healthy finalization through shared per-item mutation gates
- make dynamic repair reservations commit only after urgent state is durably saved
- use a database context factory for scheduling and release failed reservations for retry
- preserve newer streaming evidence during health finalization
- publish health status only after persistence succeeds

## Validation
- `dotnet build backend/NzbWebDAV.csproj -c Release --no-restore -p:RunAnalyzers=false` passed
- focused backend tests passed: 138 tests, including stream, middleware, health, tracker, and race regressions

## Issue Link
- `Refs #1388` — the verified scheduling/finalization failure-loss mechanisms are addressed.
- The reporter-specific first-article reproduction remains unconfirmed, so this PR intentionally does not close #1388 or claim the entire report resolved.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved repair scheduling reliability by preventing duplicate, conflicting, stale, or failed repair requests.
  - Health checks now preserve urgent repairs and newer streaming failures during concurrent events.
  - Health-check status updates are recorded only after successful persistence.
  - Improved consistency when repairs and health checks run simultaneously, reducing outdated results overriding urgent actions.

- **Tests**
  - Added coverage for concurrent urgent repairs, outdated failure snapshots, serialized failure handling, and repair scheduling edge cases.


hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai review

hoivikaj:
@coderabbitai full review

PR #1409, opened 2026-09-13
fix(nntp): bound fresh connection acquisition

### LINKED ISSUE #1396: Playback stalls while a provider's fresh connections hang, and the provider stays in selection
### Before submitting

### InfiniDysk version

1.3.0 (ghcr.io/infinidysk/infinidysk:v1.3.0)

### Installation method

Official Docker image

### Affected area

Usenet provider connection

### What happened?

A direct-play stream buffered repeatedly over fourteen minutes while one of three pooled
providers was intermittently hanging on connection open. `SegmentFetches` looks clean for the
whole window: zero misses, zero errors, zero retries, and that provider's BODY responses
under 1 s in those minutes. The breaker never opened, and the only log lines are the pre-warm
warnings below.

The stall is in connection acquisition. `MetricEvents` `streaming/body` `pool-wait` for that
provider peaked at 16,740 ms, 14,860 ms, 16,083 ms and 8,077 ms in four separate minutes,
against 148 ms for the other two across the same minutes. One sample accounts for almost the
whole `SumMs` in each of those minutes, so a single fetch at the head of the fill sat for the
connect ceiling while the rest waited on it:
`MultiSegmentStream.DownloadPipelinedSegments` awaits each batch before issuing the next.

The first read started against a cold pool: the provider had served nothing since 21:05, and
with the stock idle timeout and no warm floor the sweeper empties a pool inside about a minute
and a half. The later stalls each began when a fetch found no idle connection on that provider.
Either way the head fetch needed a handshake. The read-start warm-up
(`NzbFileStream.StartConnectionPrewarm` into `ConnectionPool.WarmToAsync`) runs handshakes
three at a time per provider through the same three-permit `_handshakeGate` a streaming caller
needs for a fresh connection, and each failed handshake arms the backoff
(`GetHandshakeFailureBackoffMs`, 500 ms doubling per consecutive failure, capped at 60 s) that
the next borrower waits out.

Nothing bounds that wait. The per-segment deadline is armed after acquisition:
`DecodedBodiesAsync` awaits `AcquireConnectionLockAsync` with the caller's token and only then
creates the `CancelAfter(PerSegmentTimeout)` source, so the deadline covers the BODY and not
the wait for a connection, while `usenet.streaming-read-timeout-seconds` covers acquisition
only until `GetAndHeadHandlerPatch` sets it to infinite at the start of body copy. Failover
cannot cover it either, because failover is driven by exceptions and a wait throws none.

The same fault fifteen hours earlier, on the health sweep, was handled correctly: a fetch-path
failure reaches `RecordConnectionAcquisitionFailure`, which trips the breaker at once when the
pool is cold, and the provider walk then logs the fallback and moves on. That branch was never
reached during playback. The warm-up path records nothing: `WarmToCoreAsync` catches the
exception, logs a warning and returns. By the time each failure was logged the provider was
completing BODYs on open sockets, 8 to 38 in the surrounding twenty seconds, so a fetch-path
failure at the same instant would have been a `RecordFailure` sample that the next
`RecordSuccess` clears. Neither path can bench a provider that serves on open sockets and
hangs on new ones.

### Steps to reproduce

1. Configure two or more pooled providers with the stock
   `usenet.streaming-segment-timeout-seconds` and `usenet.read-start-warmup.enabled`.
2. Make fresh connects to one of them hang past `UsenetStreamingClient.ConnectTimeout` without
   refusing, so its open connections keep working. The hang has to be intermittent: a permanent
   one also fails the fetch path, which trips the breaker at once on a cold pool.
3. Let the idle sweep empty that provider's pool, then open a range of 8 MiB or more so the
   read-start warm-up fires.
4. The first byte arrives about 15 s later. `MetricEvents` records one `streaming/body`
   `pool-wait` sample near the ceiling, the log shows only the pre-warm warning, and the
   breaker's trip count does not move. Each later stream open repeats it, idle gap or not.

### Expected behavior

A hung connection open should fail the fetch over inside a bounded wait and count against the
provider, and that bound wants its own setting rather than a share of the per-segment budget.
`UsenetStreamingClient.ConnectTimeout` already bounds one handshake, and its comment already
states the intent: `short enough that three stuck handshakes cannot pin the pool forever`. It
is a hard-coded 15 s with no config key, it does not bound a caller queued behind those three,
and the failure backoff stacks on top of it. A configurable deadline over the
connection-open step inside `AcquireConnectionLockAsync`, so it covers the first-segment and
seek-probe reads as well as the pipelined batch, with expiry classified as a retryable provider
wait, would let the walk move on the way the existing streaming-timeout catch already does for a
slow BODY.

Three waits are involved and they want different numbers, which is why one knob will not serve.
A permit wait on a saturated pool legitimately runs to several seconds and must not be cut
short. A BODY transfer has `usenet.streaming-segment-timeout-seconds`. Opening a connection has no setting
of its own: a handshake on the other two providers never exceeded 148 ms across those minutes,
so one still outstanding after a second or two is already anomalous, and tuning that down
should not shorten a BODY transfer. Buffering cannot absorb the difference either: these stalls land on
the head fetch of a newly opened stream, where the task window is empty and nothing has been
read ahead for the consumer to spend, so nothing on the server side shortens what the viewer
waits.

A connect failure should reach the breaker from either path and survive the next BODY success.
`RecordConnectionAcquisitionFailure` already separates the cold-pool case; the live case has the
same evidence, an open that exceeded the bound while the peers open in under 150 ms, and warrants
the same first-occurrence treatment rather than a share of a rate window, otherwise every expiry
before the trip costs the viewer another bound. Either would let a provider that serves on open
sockets but cannot open new ones leave rotation.

Together the two would make an account that intermittently hangs on connect safe to keep
pooled. The first affected stream open costs a bounded wait, the breaker then keeps later opens
off that provider until it recovers, and the pool absorbs the fault instead of the viewer.

That `usenet.streaming-priority` reaches the connection gate but not `_handshakeGate` is
secondary: a held permit cannot be reclaimed, so a deadline and a breaker signal are what bound
the stall.

```text
[21:09:43 WRN] NNTP connection pre-warm stopped for <host>. Reason: Connection to <host>:563 timed out after 15s.
[21:10:01 WRN] NNTP connection pre-warm stopped for <host>. Reason: Connection to <host>:563 timed out after 15s.
[21:14:53 WRN] NNTP connection pre-warm stopped for <host>. Reason: Connection to <host>:563 timed out after 15s.
```

`MetricEvents`, `Kind=latency`, `RefId=streaming/body`, `Tag2=pool-wait`, that provider, as
Num / SumMs / MaxMs.

```text
21:09   17 / 18855 / 16740
21:14   13 / 14901 / 14860
21:21   13 / 16083 / 16083
21:23   75 /  9557 /  8077
```

The handled case on the fetch path, the same day:

```text
[05:48:26 WRN] Provider <host> tripped (connection failure (get-connection-CouldNotConnectToUsenetException: Connection to <host>:563 timed out after 15s.)). Pool live=0, idle=0, active=0. Skipping for 60s.
[05:48:26 INF] Provider <host> error: Connection to <host>:563 timed out after 15s.. Falling back to <other-host>
```

### Environment

OS: Ubuntu Server 24.04 LTS
Architecture: x86_64
WebDAV client: rclone 1.75.0 with --vfs-cache-mode=full
Consumer: Plex
Integration: Radarr 6.3.0 through the SABnzbd-compatible API

### Additional context

#351 introduced the 15 s connect ceiling, and its problem statement already described three
hung connects freezing connection creation for a provider while the breaker stayed silent. The
ceiling bounds each hang; it does not make the hang visible to the breaker or bounded for a
playback read. #1185 is an open validation task for the same missing acquisition deadline on
the queue side.

The provider-side cause is unknown, intermittent, and not one account's property. Over three
months, with roughly 2,300 hourly test connections per provider alongside normal traffic, three
of six accounts hung between one and four times each, one of them at the AUTHINFO stage after
TLS completed. The three that hung read the UsenetExpress and NetNews backbones; the three that
never hung are all Omicron, so a natural reproduction is more likely against the first two. The
injected hang in the steps above needs no particular backbone. Any provider can present this,
so the pool has to survive it without the stream noticing.

hoivikaj:
## Summary
- add `usenet.connection-open-timeout-seconds` with a 1-15 second bound and 15 second default
- bound fresh connection handshake admission, creation capacity, replacement pacing, and factory time
- apply the same configurable deadline to warm/prewarm connection creation
- track non-cooperative factory completion so late sockets are disposed and permits are released exactly once
- expose typed retryable timeout failures with provider and phase context
- report warm factory failures to the provider breaker while preserving connection-limit shrink behavior
- require a genuine fresh half-open connection before a strong fresh-open circuit can recover
- thread probe leases and fresh/reused proof through singular, batch, and pipelined paths
- add failed acquisition latency recording, effective-config manifest coverage, managed Settings control, documentation, and regression coverage
- preserve normal idle reuse and caller/shutdown cancellation behavior

Closes #1396

## Validation
- `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --no-restore --filter 'FullyQualifiedName~ConnectionPool|FullyQualifiedName~ProviderCircuitBreakerConnectionFailureTests|FullyQualifiedName~ProviderCircuitBreakerHalfOpenTests|FullyQualifiedName~ConnectionAcquisition'` (89 passed)
- `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --no-restore --filter 'FullyQualifiedName~ConnectionPoolWarmConnectionTests|FullyQualifiedName~ConnectionPoolPriorityTests'` (21 passed)
- `cd frontend && npx vitest run app/routes/settings/streaming/streaming.test.ts` (9 passed)
- `dotnet build backend/NzbWebDAV.csproj -c Release --no-restore` passed
- editor diagnostics and `git diff --check` clean

## Summary by CodeRabbit

- **New Features**
  - Added a configurable fresh-connection open timeout, defaulting to 15 seconds and adjustable from 1–15 seconds.
  - Added the “Fresh Connection Open Timeout” option to streaming settings and effective configuration details.
  - Improved provider recovery by requiring fresh connections when connection establishment failures indicate they are needed.

- **Bug Fixes**
  - Improved handling of timed-out connection attempts and warm-connection failures.
  - Pool-wait latency is now recorded consistently, including failure scenarios.

- **Documentation**
  - Documented timeout behavior, scope, and configuration effects.


loambit:
@hoivikaj Saw the note that the breaker portion is deferred, is that tracked anywhere I can follow? Also looks like the pre-warm path still uses the old 15s timeout rather than the new setting.

hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai full review

loambit:
@hoivikaj appreciate the hard work on this!

PR #1435, opened 2026-09-15
fix(metrics): quarantine a corrupt metrics database and back off failed flushes

job-bot-ai:
## Summary

A corrupt `metrics.sqlite` currently degrades the whole backend until an operator finds it. This PR makes the metrics database self-healing and stops the flush loop from spinning when the database is unusable.

**What happened on a real host** (Unraid, `metrics.sqlite` on a FUSE-backed appdata share; the share's FUSE daemon crashed mid-write):

- `PRAGMA quick_check` on the 2 GB file: `Tree 14 page 418537 cell 193: 2nd reference to page 91771` (and more).
- The backend kept starting (`__EFMigrationsHistory` was readable), so nothing surfaced at startup — `DatabaseIntegrityCheck` deliberately skips the metrics DB.
- `MetricsWriter` retried its flush in a hot loop: `WaitForFlushAsync` returns immediately once any queue holds 1000 rows, the flush threw `SQLITE_CORRUPT`, the batch was requeued, repeat. ~6,300 `MetricsWriter flush failed` warnings per 10 minutes, one core pegged, and the backend's WebDAV first-byte latency rose to seconds (`Slow WebDAV metadata request … DurationMs=7095`), which slowed every arr import behind it.
- `MetricsRollupService` and the arr health poller failed on every tick as well.

## Changes

- **`MetricsDatabaseRecovery`** (new): runs `PRAGMA quick_check` on the metrics database; if it reports findings, or the file throws `SQLITE_CORRUPT` (11) / `SQLITE_NOTADB` (26), the file and its `-wal`/`-shm`/`-journal` sidecars are moved to `metrics.sqlite.corrupt-<utc timestamp>` (kept for inspection) and one `Error` is logged with the findings. The context's connection is closed first and pooled connections are cleared, so the next open — the existing `MigrateAsync` — creates a fresh database. Never throws, never blocks startup. Only the metrics database is treated this way; the main database keeps its diagnostic-only behaviour because it is not disposable.
- **Startup**: `StartupDatabaseMigrator` and the `--db-migration` maintenance path in `Program.cs` call the recovery before anything touches the metrics database.
- **`MetricsWriter`**:
  - counts consecutive flush failures and backs off exponentially between attempts (1 s, 2 s, 4 s … capped at 60 s), reset on the next successful flush. A broken database now costs one warning per minute instead of ten per second.
  - when the failure is `SQLITE_CORRUPT` / `SQLITE_NOTADB`, quarantines and recreates the database at runtime (rate-limited to once per 10 minutes) so metrics resume without a restart. Queued rows stay in memory and land in the new file on the next flush.
- Tests: `MetricsDatabaseRecoveryTests` (missing file, healthy file, not-a-database file, malformed page → quarantine + recreate, error-code classification) and additions to `MetricsWriterTests` (failure counter, backoff table, reset on success).

## Notes for reviewers

- Quarantine is a rename, not a delete: a 2 GB corrupt file stays on disk until the operator removes it. I preferred keeping the evidence over silently freeing space; happy to switch to delete-after-N-days if you'd rather.
- `File.Move(…, overwrite: true)` on the sidecars only matters if a previous quarantine within the same second left files behind.
- The root cause on my host was environmental (SQLite on a FUSE share that crashed; the file has since moved to a local filesystem). The PR is about containing the blast radius when it happens anyway.

## Test plan

- [x] `dotnet build backend/NzbWebDAV.csproj`
- [x] `dotnet test tests/NzbWebDAV.Tests --filter "FullyQualifiedName~Metrics"`
- [x] On the affected host: with the corrupt file in place the backend now logs one `Metrics database is corrupt and was quarantined to …` at startup, creates a fresh `metrics.sqlite`, and `MetricsWriter` flushes normally afterwards.

## Summary by CodeRabbit

- **Reliability Improvements**
  - Corrupted metrics databases are detected during startup and safely quarantined so a fresh database can be created.
  - Failed metrics flushes retry with exponentially increasing delays, capped at one minute.
  - Successful flushes reset the failure count and retry delay.
  - Busy, locked, and disk-related errors are not treated as database corruption.

- **Tests**
  - Added coverage for database recovery, corruption detection, quarantine behavior, and flush retry handling.

## Update: root cause found (after opening this PR)

The corruption on my host was not the filesystem crash after all. A host-side maintenance script was opening the *live* `metrics.sqlite` with the `sqlite3` CLI every 10 minutes to prune old rows and run `PRAGMA wal_checkpoint(TRUNCATE)` — through a FUSE path (`/mnt/user/…`, Unraid's shfs) while the backend held the same file through a direct bind mount. FUSE and the underlying filesystem do not share POSIX lock state, so each side believed it was the only user: the external DELETEs interleaved with the backend's writes (→ malformed page tree), and the external checkpoint+close unlinked the WAL/SHM the backend had memory-mapped (→ `SQLite Error 10: disk I/O error` on every subsequent flush; the backend's fd pointed at `metrics.sqlite-wal (deleted)`).

That script is gone on my side. It doesn't change what this PR does — it's still the case that a damaged metrics database currently degrades the whole backend until an operator intervenes — but it does suggest a follow-up worth considering: on `SQLITE_IOERR` (10) the writer could clear the connection pool and reopen, which would recover from an externally deleted WAL/SHM without a restart. I left that out here because I couldn't build a deterministic test for it.


hoivikaj:
@coderabbitai full review

PR #1453, opened 2026-09-16
fix(queue): lazy RAR mounting no longer drops extra files hidden behind the first archive member

### LINKED ISSUE #1413: Lazy RAR can accept multi-file archives and silently omit later members
### Before submitting

### InfiniDysk version

1.3.0; also present on main@11d587beb8523ed4ca95d04369dc59b4bff2871a

### Installation method

Official Docker image

### Affected area

Queue, NZB processing, or health checks

### What happened?

<p>A source-level review found that <code>LazyRarProcessor</code> can accept a valid multi-file RAR archive as a lazy single-member archive and omit later members from the resulting mount.</p>

<p>The lazy processor is designed to return <code>null</code> when an archive does not match the supported lazy shape, allowing the ordinary eager RAR processor to handle that archive instead.</p>

<p>However, the current implementation explicitly notes that the former "must be a single-file archive" check was removed because proving that condition required walking beyond the first file header.</p>

<p>The replacement is a size heuristic:</p>

<pre><code>selected first member size >= 70% of total known NZB/archive size</code></pre>

<p>That ratio is only checked when every RAR volume has a known <code>FileSize</code>. When one or more sizes are unavailable, the ratio check is skipped.</p>

<p>Even when the ratio is available, it does not prove that the archive contains only one file.</p>

<p>The first physical volume is parsed using <code>RarUtil.ReadHeadersUntilFirstFileAsync()</code>, which deliberately stops as soon as it yields the first non-directory file header.</p>

<p>For a supported stored/m0 single-volume archive such as:</p>

<pre><code>archive.rar
├── Episode01.mkv   8 GiB
└── Episode02.mkv   2 GiB</code></pre>

<p>the first member represents 80% of the archive and therefore satisfies the current 70% heuristic.</p>

<p>If <code>Episode01.mkv</code> is a complete, non-split stored member, <code>ValidateContinuationChainAsync()</code> can validate that first member's packed size and return success without proving that another non-directory file header follows it in the same volume.</p>

<p><code>LazyRarProcessor.ProcessAsync()</code> then returns one lazy result for <code>Episode01.mkv</code>.</p>

<p>Because lazy processing succeeded for that resolved archive set, the eager RAR processor does not enumerate the archive's complete member list. <code>Episode02.mkv</code> therefore has no representation in the completed mount.</p>

<p>The normal final media validator does not establish archive completeness; the presence of one importable mounted media item is sufficient for that validator.</p>

<p>The same removed single-file check and 70% replacement heuristic are present in <code>v1.3.0</code> and current <code>main</code>.</p>

### Steps to reproduce

<ol>
<li>Create a deterministic valid RAR4 or RAR5 archive using supported stored/m0 entries.</li>

<li>Place at least two non-directory file members in one physical archive volume, with the first member large enough to exceed the current 70% threshold. For example:
<pre><code>Episode01.mkv = 80 MiB
Episode02.mkv = 20 MiB</code></pre>
The exact sizes are not important as long as the first member exceeds the threshold and both entries are structurally valid.</li>

<li>Represent the archive as a <code>GetFileInfosStep.FileInfo</code> input suitable for <code>LazyRarProcessor</code>, with lazy RAR parsing enabled.</li>

<li>Supply a correct physical <code>FileSize</code> so the ratio check executes.</li>

<li>Call <code>LazyRarProcessor.ProcessAsync()</code>.</li>

<li>Observe that <code>ReadHeadersUntilFirstFileAsync()</code> stops after the first member header.</li>

<li>Observe that the first member passes the 70% ratio check.</li>

<li>Because there are no trailing RAR volumes and the first member is not split, observe that continuation validation can succeed based on that member's size.</li>

<li>Observe that <code>ProcessAsync()</code> returns a non-null lazy result representing only <code>Episode01.mkv</code>.</li>

<li>Run that result through the normal archive-set queue path. Because lazy processing succeeded, eager enumeration does not discover <code>Episode02.mkv</code>.</li>
</ol>

<p>Repeat the test with PAR2-derived <code>FileSize</code> unavailable. In that case the 70% heuristic is skipped entirely, but the missing structural single-member proof remains.</p>

### Expected behavior

<p>Lazy RAR processing should succeed only when InfiniDysk has structurally established that the selected member is the only relevant non-directory file member in that archive set.</p>

<p>If another non-directory file member follows the selected member, <code>LazyRarProcessor.ProcessAsync()</code> should return <code>null</code> and allow the existing eager RAR processor to enumerate and represent all members.</p>

<p>If single-member eligibility cannot be proved safely, the implementation should also fall back to eager processing rather than infer completeness from file-size ratio.</p>

<p>The existing 70% test may remain as an inexpensive rejection heuristic, but it should not be the correctness proof for single-member eligibility.</p>

<p>The change should preserve successful lazy behavior for genuine single-member RARs, current multipart continuation validation, encryption behavior, retryable transport-error classification and cancellation semantics.</p>

<p>No runtime support pack or provider-specific log is required for this deterministic source-level reproduction. No NZB contents or credentials are included.</p>

<p>Relevant source:</p>

<ul>
<li><a href="https://github.com/infinidysk/infinidysk/blob/11d587beb8523ed4ca95d04369dc59b4bff2871a/backend/Queue/FileProcessors/LazyRarProcessor.cs"><code>backend/Queue/FileProcessors/LazyRarProcessor.cs</code></a> — documents removal of the explicit single-file check, reads only through the first file header and substitutes the 70% heuristic.</li>

<li><a href="https://github.com/infinidysk/infinidysk/blob/11d587beb8523ed4ca95d04369dc59b4bff2871a/backend/Utils/RarUtil.cs"><code>backend/Utils/RarUtil.cs</code></a> — <code>ReadHeadersUntilFirstFileAsync()</code> intentionally stops after the first non-directory file header.</li>

<li><a href="https://github.com/infinidysk/infinidysk/blob/11d587beb8523ed4ca95d04369dc59b4bff2871a/backend/Queue/QueueItemProcessor.cs"><code>backend/Queue/QueueItemProcessor.cs</code></a> — successful lazy processing is used instead of eager processing for that resolved RAR set.</li>

<li><a href="https://github.com/infinidysk/infinidysk/blob/11d587beb8523ed4ca95d04369dc59b4bff2871a/backend/Queue/PostProcessors/EnsureImportableMediaValidator.cs"><code>backend/Queue/PostProcessors/EnsureImportableMediaValidator.cs</code></a> — final media validation proves that at least one mounted media item exists, not that every archive member was represented.</li>

<li><a href="https://github.com/infinidysk/infinidysk/blob/11d587beb8523ed4ca95d04369dc59b4bff2871a/tests/SharpCompress.Tests/Rar/RarHeaderFactoryDeferredSkipTests.cs"><code>tests/SharpCompress.Tests/Rar/RarHeaderFactoryDeferredSkipTests.cs</code></a> — establishes the in-tree deferred header-skip behavior relevant to a metadata-only completeness check.</li>
</ul>

### Environment

<pre><code>Repository: infinidysk/infinidysk
Affected versions verified in source: v1.3.0 and main
Current main commit: 11d587beb8523ed4ca95d04369dc59b4bff2871a
Investigation: deterministic source-level analysis
Lazy RAR parsing: enabled
RAR member compression: stored/m0
OS: not material to reproduction
Architecture: not material to reproduction
Usenet provider: not required
WebDAV client: not required</code></pre>

### Additional context

<p><strong>Suggested fix direction</strong></p>

<p>Keep multi-member support in the existing eager RAR path rather than expanding the lazy representation.</p>

<p>Before returning a successful lazy result, add a structural single-member proof at the point where the selected member terminates:</p>

<ol>
<li>Parse and validate the selected member as today.</li>
<li>Advance header enumeration beyond that member in its terminating physical RAR volume.</li>
<li>Ignore directory/service metadata as appropriate.</li>
<li>If another non-directory file header is present, return <code>null</code> and use the eager processor.</li>
<li>Only return a lazy result after header enumeration establishes that no later file member exists.</li>
</ol>

<p>For a single-volume archive, the terminating volume is the first volume. For a multipart selected member, the structural check belongs in the final volume where that member's <code>IsSplitAfter</code> chain ends.</p>

<p>The in-tree SharpCompress implementation already uses deferred packed-data skipping, so this can remain a header/metadata validation rather than archive extraction.</p>

<p>If advancing through the terminating volume requires an accurate physical stream length and PAR2 size metadata is unavailable, use the existing accurate size-probe path rather than treating an estimated decoded length as authoritative.</p>

<p><strong>Recommended regression coverage</strong></p>

<ul>
<li>Single-volume archive with first member &gt;70% plus a second file member → lazy returns <code>null</code>.</li>
<li>Same archive shape without PAR2-derived <code>FileSize</code> → lazy returns <code>null</code>.</li>
<li>Multipart selected member ending in the final RAR volume with another file member after it → lazy returns <code>null</code>.</li>
<li>True single-member single-volume RAR → current lazy success remains.</li>
<li>True single-member multipart RAR → current lazy success remains.</li>
<li>Directory/service headers after the selected member do not produce a false multi-file rejection.</li>
<li>Transient provider errors encountered during the additional structural proof remain retryable.</li>
<li>Cancellation continues to propagate normally.</li>
<li>Existing supported encrypted single-member behavior remains unchanged.</li>
</ul>

<p>No database migration, new configuration option or UI change should be required.</p>

hoivikaj:
## Summary
- Prove lazy RAR eligibility by enumerating the terminating volume for additional non-directory members.
- Always use the terminating volume's physical decoded length for the proof, preserving retryable and cancellation behavior.
- Add multi-member RAR4 fixtures and coverage for single/multipart fallback, directories, stale sizes, retryable failures, and cancellation.

Closes #1413

## Test plan
- `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --filter 'FullyQualifiedName~LazyRarProcessorTests|FullyQualifiedName~QueueRarTimeoutTests|FullyQualifiedName~RarUtil' --no-restore`
- Result: 40 passed.

Non-breaking; no migration, configuration, or UI changes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved archive validation before lazy mounting across single- and multi-volume archives.
  - Archives containing multiple files are no longer incorrectly mounted as single-file archives.
  - Directory entries no longer cause valid single-file archives to be rejected.
  - Transient validation failures now support retry handling, with cancellation handled correctly.

- **Tests**
  - Expanded coverage for multi-member, multi-volume, directory, retry, and cancellation scenarios.


hoivikaj:
@coderabbitai full review

hoivikaj:
@coderabbitai full review

PR #1460, opened 2026-09-17
fix(usenet): allow imports when yEnc headers omit the total count

hoivikaj:
## Summary

Fix a reproduced import regression in the yEnc count validation introduced in 1.4.0: a header with no `total` field was parsed as `TotalParts = 0`, then rejected against the NZB's multipart segment count.

- Preserve total-field presence in `UsenetYencHeader.HasTotalParts`: `false` means parser-confirmed omission, `true` means the field was supplied, and `null` preserves conservative behavior for older cached or manually constructed headers.
- Skip the unavailable count comparison only when omission is confirmed and the parsed count is zero. Declared positive mismatches remain rejected, even if combined with an inconsistent omission flag.
- Preserve rejection of explicit zero and malformed totals for multipart NZBs. The existing single-segment zero-total allowance is unchanged.
- Preserve omission metadata through cache serialization; legacy cache entries without this field retain the existing validation/refetch behavior.

No offset-based identity heuristic is introduced. CRC checks, byte-range checks, provider fallback, stream disposal, and connection lifecycle code are unchanged. Omission means the count check cannot establish a mismatch; it is not proof of article identity or payload integrity.

## Reproduction and Local Verification

A local instance using diagnostic PR #1459 reproduced three imports failing with `part 66/0 for a file with 66 parts`, alongside three successful imports. A bounded, authorized TLS BODY-header probe of one rejected article confirmed the raw `=ybegin` had **no `total` field**, rather than an explicit or malformed zero. It declared part 66, file size 50,000,000, and `=ypart begin=49920001 end=50000000`.

After publishing this fix and restarting the backend against the same local configuration/database, exactly one previously failed NZB was retried through the existing history retry API. It **completed in 2 seconds**, and history confirmed `Completed`. That retry logged no count rejection, missing-RAR error, or CRC error.

This verifies that import path, not full playback, complete payload CRC verification, or SABnzbd repair equivalence. Other reports involving positive count mismatches, missing RAR volumes, or Arr failure provenance are not claimed fixed by this PR.

Only anonymized observations are included here. No provider credentials, article IDs, filenames, NZBs, raw logs, support bundles, or local investigation plans are committed or attached.

## Test Plan

Passed locally: **82 focused tests**, no failures or skips:

```sh
dotnet test tests/UsenetSharp.Tests/UsenetSharp.Tests.csproj -c Release --filter 'FullyQualifiedName~ParseYencHeaders' --no-restore
# 9 passed

dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj -c Release --filter 'FullyQualifiedName~MultiProviderNntpClientYencValidationTests|FullyQualifiedName~HeaderCachingNntpClientTests|FullyQualifiedName~SegmentCacheNntpClientTests' --no-restore
# 73 passed
```

Coverage includes omitted, explicit-zero, empty, malformed, negative, matching, and mismatched totals; size probing; sequential and pipelined streaming; fallback; and confirmed-omission versus legacy-header cache hydration. Release publish also passed. Broad CI-covered checks are left to PR CI per repository guidance.

## Scope

Independent behavioral fix based on `main`; does not include or change the purpose of diagnostic PR #1459. No configuration options, setup-wizard changes, database migrations, or supplied-evidence edits. No applicable root PR template was found.


hoivikaj:
Addressed the bare-`total` finding in [Copilot's review](https://github.com/infinidysk/infinidysk/pull/1460#pullrequestreview-5236895720) with commit 2a663487.

A token exactly equal to `total` without `=` now records `HasTotalParts = true` and `TotalParts = 0`, so it is treated as a malformed declaration rather than confirmed omission. The existing multipart guard therefore rejects it; genuinely omitted totals remain accepted.

Added a regression to `ParseYencHeaders_TotalPresence_DistinguishesOmittedFromInvalid`. Verified red/green: it failed before the parser change (`HasTotalParts` was false), and all **10 parser tests** pass afterward:

```sh
dotnet test tests/UsenetSharp.Tests/UsenetSharp.Tests.csproj -c Release --filter 'FullyQualifiedName~ParseYencHeaders' --no-restore
```

This follow-up changes only the parser and its regression case. The prior 73 backend/cache tests and successful live import retry are documented above; those were performed on the original fix, not rerun for this parser-only follow-up.

PR #1468, opened 2026-09-18
fix(queue): report unsupported 7z compression instead of missing byte ranges

hoivikaj:
## Summary
- Defer 7z packed-range lookup until the processor has checked archive capabilities.
- Compressed archives now produce the existing actionable unsupported-compression failure instead of a misleading missing-packed-range exception.
- Preserve packed-range validation for supported stored archives; no new decompression support or yEnc-validation changes.

## Reproduction
Two privately supplied NZBs were tested against clean committed main at d55e941e, independent of local work in progress. Both failed in archive handling, with no yEnc mismatch rejections or PAR2 metadata-discovery warnings:
- RAR sample: existing unsupported compression method restriction (only stored/m0 supported).
- 7z sample: missing packed-byte-range exception before the compression capability check.

With this change, the same 7z sample reports `Only uncompressed 7z files are supported.` The sample still cannot import: this change corrects error classification, not archive support. Compressed archive import/seek support requires separate design work.

## Validation
- 20 focused SevenZipUtilTests and SevenZipProcessorTests passed.
- Added coverage using existing LZMA, solid, AES+LZMA, and AES+LZMA2 fixtures.
- Release publish passed; no editor diagnostics; git diff whitespace checks passed.
- Private NZBs, provider credentials, and logs remain local and are not included.
- Isolated validation backend stopped; existing checkout edits preserved.

No configuration changes, setup-wizard impact, or migrations.

## Summary by CodeRabbit

- **Bug Fixes**
  - Archive entries now expose their compression type for LZMA, solid, and AES-encrypted 7z archives even when packed byte-range information is unavailable.
  - Packed byte-range details are evaluated when requested and report an error if the archive entry does not provide that information.


hoivikaj:
@coderabbitai full review

PR #1494, opened 2026-09-19
fix(health): flag unlinked files and allow needs-attention removal with read-only WebDAV

hoivikaj:
## Summary
- Add an orange **Not library linked** pill to Needs attention rows with the missing imported symlink or `.strm` reason, on desktop and mobile.
- Allow preview and deletion of current Needs attention files when WebDAV Enforce Read-Only is enabled. Validate the selected health-result ID against the file and latest health status; reject stale results, mismatched files, and directories. Preserve read-only user permissions and existing deletion protections.
- Remove the indexer Search action from these rows.

## Validation
- 43 focused frontend tests passed (health table and route component).
- 30 DeleteWebdavItemControllerTests passed, including read-only exception and rejection cases.
- Runtime OpenAPI contract check passed; no contract diff.
- Scoped frontend ESLint passed; editor diagnostics and design scan reported no issues.
- Real table component checked with fixture data at 390px and 1440px: orange badge fits, no horizontal overflow, no search link.
- Live authenticated end-to-end deletion was not exercised; the preview server required sign-in. Backend deletion and frontend confirmation were tested independently.

## Configuration
No new configuration or persisted setting; no setup-wizard change is needed.

## Summary by CodeRabbit

- **New Features**
  - Health-based file deletion now verifies that the selected attention result is still current before proceeding.
  - Items no longer matching the selected result are rejected with a conflict message.
  - “Needs attention” entries can display a **Not library linked** status badge when applicable.

- **Changes**
  - Health-based deletions can proceed even when WebDAV is read-only, while ordinary deletions remain restricted.
  - Removed the Search action from health-history entries; deletion remains available independently of re-check status.


hoivikaj:
Addressed the remaining CodeRabbit outside-diff finding in b7925e8: the health-result mock now has a concrete GUID, and the test asserts that exact GUID is sent in both the preview query and deletion form. Focused validation passes (17 frontend tests, 7 backend tests).

PR #1497, opened 2026-09-19
fix(nntp): accept yEnc articles with a decoy total when part and offset match

scottrobertson:
Some obfuscated uploads put random numbers in the yEnc `total=`, `size=` and `name=` fields while `part=` and the `=ypart` byte range stay correct. Since #1357 the streaming identity check compares only `total=` against the NZB segment count, so every segment of those posts is rejected as a wrong article, the stream gets gap filled, and the urgent repair path then deletes the file from Sonarr or Radarr and blocklists the release. The data on Usenet is fine. Health checks never flag it because they STAT and the article does exist.

A real example, from a segment that is 768,000 bytes on the wire:

```
=ybegin part=1 total=781 line=128 size=11813576 name=W1uv7rY043t9p9IV7RgYVt0i1
```

That article decodes with a matching CRC to the start of the RAR volume the NZB expects. It isn't limited to one uploader or release type either, the same headers turn up across a wide range of posts at both 1080p and 2160p.

The total check stays as it is. When it disagrees, the article is now also accepted if its part number matches the segment position that was requested and its `=ypart` offset is where that segment should start. The last segment only has to match on part number, since it can be shorter. A wrong article still fails, because its part number or offset points somewhere else.

Tested against a genuine message-id collision as well, where one provider serves unrelated articles for 115 of 2,744 segments in a release. Those are still rejected and the reads fall through to the second provider.

A note on how this came about: I spent hours debugging this trying to work out why whole seasons were being flagged and deleted, and the code in this PR was written by Claude once we'd narrowed it down. I don't know C#, so excuse the code if it isn't how you'd write it. It's mainly here to show the issue and a fix that I've verified works for me. Feel free to rework it however you like.


hoivikaj:
We think this was already addressed by #1461

PR #1506, opened 2026-09-20
fix(usenet): prevent queued connection waits from stalling provider failover

### LINKED ISSUE #1501: A benched provider keeps its queued requests and pre-factory waits have no bound
### Before submitting

### InfiniDysk version

1.4.2 (ghcr.io/infinidysk/infinidysk:v1.4.2). Code read on main at d04075a6, which includes #1495.

### Installation method

Official Docker image

### Affected area

Usenet provider connection

### What happened?

Hopefully the last issue on the "connection open hang" problem, when a provider is benched by the circuit breaker all of its waiting connections should be purged as well (with some exceptions).

One pooled provider's opens hang past the open budget a few times per day. The logs below were taken with that budget at 3 s. The breaker tripped it ten times in seven days. One trip came from the read-start pre-warm with `Pool live=36, idle=36, active=0`. `RecordConnectionFailure` trips on one factory timeout and 36 working connections did not count.

The trip gates admission only. `TryAdmitBeforeAcquisition` runs before the pool is touched and `ProviderCircuitBreaker` exposes no cancel or drain. Requests already past admission keep their place on the pool gate and the handshake gate. Each one later runs the factory and spends the open budget against a benched provider. The log below shows two queued requests failing one second after a trip. Under #1495 the same requests wait with no deadline and then run the factory.

Nothing bounds the wait before the factory. `GetConnectionLockCoreAsync` links only the caller token and pool disposal into every wait before the factory. #1495 arms the open deadline in the `Factory` phase alone. A playback request therefore waits until its own caller cancels it. A queue worker waits for the five-minute stuck watchdog. #1185 asks this question and is still open.

`AcquireConnectionLockAsync` already bounds provider admission with `TransferAdmissionFailoverContext.WaitTimeout` of 15 s when another provider has capacity. Expiry fails over to the next provider. That bound stops at provider admission and nothing after it has one. Retries re-verify admission, so only requests already past it fall through.

`GetConnectionOpenTimeout` clamps `usenet.connection-open-timeout-seconds` to 1..15 with a default of 15. Both values predate #1495 and sized a budget that included queue waits. The budget now covers the socket open only. Opens measured 0.04 to 0.10 s from this host. A 15 s default now holds a playback fetch on one hung socket.

### Steps to reproduce

1. Configure pooled providers A and B with the open timeout left at 15.
2. Point B's host at a TCP listener that accepts and never completes the TLS handshake.
3. Start twenty BODY requests against B with no idle connections in its pool.
4. The first factory timeout trips B. Each queued request still enters the factory and spends 15 s before failing over.
5. Hold a request on B's pool gate with all permits taken. The request waits until its caller cancels it.

### Expected behavior

A trip stops the work behind it. When `Trip` runs, the pool fails every waiter that has not started a factory. Those waiters fail with `CircuitAdmissionRejectedException`, the retryable error the client already uses for refused admission. A trip taken with idle connections in the pool leaves those connections serving.

Every pre-factory wait carries a deadline that a healthy sibling can absorb. The `TransferAdmissionFailoverContext` pattern extends to the pool gate and every wait after it. Expiry classifies as a retryable provider wait and not as provider damage. That resolves #1185.

The open-timeout clamp is re-derived for a socket-only budget. A streaming request needs a budget near real open times. A queue request can afford more.

```text
[22:10:48 WRN] Provider <host> tripped (connection failure (warm-open-timeout-phase-Factory)). Pool live=12, idle=2, active=10. Skipping for 60s.
[22:10:49 WRN] Error getting connection-lock for provider <host>. Reason: Connection opening for provider '<host>' exceeded the 3s budget during ReplacementPacing.
```

### Environment

OS: Ubuntu Server 24.04 LTS
Architecture: x86_64
WebDAV client: rclone 1.75.1 with --vfs-cache-mode=full
Consumer: Plex
Integration: Radarr 6.3.0 through the SABnzbd-compatible API

### Additional context

_No response_

hoivikaj:
Closes #1501.

## Summary

- Rotate one shared provider acquisition cancellation epoch on circuit trips.
- Bound alternative-provider BODY/ARTICLE acquisition across health admission, provider admission, pool gates, handshake capacity, creation capacity, and replacement pacing.
- Preserve started factories and active transfers after a trip; allow idle-only reuse after fresh-connection failures while keeping command-failure circuits quarantined.
- Keep wait expiry and circuit rejection retryable and health-neutral.
- Change the unset/unparsable fresh connection-open fallback from 15 seconds to 3 seconds while preserving explicit values, including 15 seconds.
- Add focused concurrency, configuration, routing, exception, settings, and allocation benchmark coverage.

The fresh-open default is shipping in 1.4.3 as a behavior correction for this new setting; it is not being treated as a breaking major-version change. No migration or persisted-data rewrite is included.

## Setup-wizard impact review

This changes the fallback of the existing advanced Streaming timeout; it adds no
configuration key or first-install requirement. Leave setup loader/form/Review
payloads and `SetupWizardService.AllowedConfigKeys` unchanged. Leave
`SetupWizardService.CurrentWizardVersion` at 1. Both Symlinks and STRM setup paths
continue to use the runtime default unless a valid existing saved/environment value
already controls it. Setup must not submit or overwrite this key. The existing
Settings `ManagedSetting` wrapper, pinning, and save-payload omission remain
authoritative. There is no restart requirement for a subsequent open, no
existing-data rewrite, and no database migration. The 15-to-3-second effective
default change is documented for the 1.4.3 release, Settings help, and versioned
documentation. Do not bump the wizard solely for this advanced transport timeout.

## Validation

- `dotnet test tests/NzbWebDAV.Tests/NzbWebDAV.Tests.csproj --no-restore`: 5,575 passed, 16 skipped.
- Focused frontend settings tests: 23 passed.
- `dotnet build backend.Benchmarks/NzbWebDAV.Benchmarks.csproj --no-restore` passed.
- `git diff --check` passed.

## Summary by CodeRabbit

- **New Features**
  - Provider connection acquisition now respects circuit-breaker status and can reuse established idle connections when new connections are unavailable.
  - Added consistent wait handling across provider admission, connection pools, and failover, with retryable errors when admission is rejected or times out.
  - Warm-up and failover behavior now respond more gracefully to provider circuit state.

- **Configuration**
  - Fresh connection open timeout now defaults to 3 seconds, with supported values from 1–15 seconds.

- **Documentation**
  - Updated streaming and connection-warming guidance to explain acquisition waits, circuit behavior, and timeout scope.


loambit:
Some things I didn't see the review agents catch:

- MultiConnectionNntpClient.AcquireConnectionLockAsync: when _connectionAdmission.TryAcquire succeeds, acquisitionWait stays null, so the OperationCanceledException catch never matches and a pool gate budget expiry escapes as a bare cancellation that RunWithConnection records as provider damage and retries with a fresh 15s budget.
- ConnectionPool.GetConnectionLockCoreAsync: the _gate.WaitAsync call sits outside any try, so an expiry of the pool-armed waiting timer is never converted to ProviderTransferAdmissionTimeoutException("PoolGate") the way the later phases are.
- ConnectionPool.GetConnectionLockCoreAsync: the handshake-gate catch calls ThrowIfLocalOpenTimeout without the acquisition-timeout check the second catch has, so a 15s budget expiry in HandshakeQueue throws ConnectionOpenTimeoutException claiming the 3s open budget was exceeded.
- Tests: nothing drives acquisitionWaitTimeout to expiry inside the pool; add cases for gate and handshake-queue expiry asserting ProviderTransferAdmissionTimeoutException with the right phase, no breaker failure recorded, and the permit released.

hoivikaj:
@coderabbitai full review

PR #1510, opened 2026-09-20
feat(warden): mint fingerprints from failed imports

job-bot-ai:
## Why

Warden's list only grows from Watchtower and playback. An install that grabs through Sonarr/Radarr therefore filters nothing for a long time — my own install had **0 fingerprints** after weeks of use, while its history held **11,137 failed imports**, of which **~8,600 failed because the articles were gone**. Those failures are exactly the evidence Warden wants, and they are already stored: every failed queue item keeps its nzb, which carries the size, poster and post date a fingerprint is made of.

## What

- **Settings → Warden → “Scan history”**. It dry-runs first and reports what it found (scanned / dead-article failures / distinct fingerprints, plus skips). Nothing is written until you click **Add to my list**.
- `POST /api/warden-import-history` behind it — dry run unless `?dryRun=false`.

### What counts as evidence

Only failures that mean *the articles are not on the servers*:

- `missing segments across all providers`
- `Missing rar volumes detected.`

Everything else a queue item can fail with — unparsable rar headers, unsupported compression, no media files, encrypted archives, CRC mismatches — means the data arrived but could not be used, which says nothing about the release being dead. Marking those would hide healthy releases from every future search, so they are deliberately excluded. On my install that exclusion drops ~960 items (631 rar-header, 195 m0, 73 no-media, 40 encrypted, 21 CRC).

Items whose nzb is no longer stored, or whose nzb has neither poster nor date, cannot be fingerprinted and are reported as skipped rather than guessed at.

## Supporting changes

- `NzbFile` now keeps the `poster` and `date` attributes; the parser discarded both. `NzbDocument.ParsePostedAt` treats junk/zero dates as unknown rather than 1970.
- `WardenStore.MarkDeadMany` writes a batch in one transaction with a reused parameter. `MarkDead` opens a connection per call, which is right for one-at-a-time verdicts but not for thousands at once.
- A scan refuses to start while one is running (409), so an impatient second click cannot double-read every blob.

## Notes for review

- The fingerprint is computed exactly as search results are: total size with the first file's poster and post date, falling back to summed segment bytes when history recorded no size.
- The scan runs inside the request like the other maintenance scans (`remove-missing-payloads/dry-run`). On ~11k items with stored nzbs it takes well under a minute here; if you'd prefer it on the websocket progress channel instead, happy to move it.
- No new config key, so no setup-wizard impact.
- Tests: `tests/NzbWebDAV.Tests/Services/WardenHistoryImporterTests.cs` (11 cases incl. dedupe, skip counts, the exclusion list and the busy guard) and `frontend/app/routes/settings/warden/warden.test.tsx` (3 cases covering preview-before-write). Contract regenerated via `scripts/export-admin-openapi.sh`.

## Summary by CodeRabbit

- **New Features**
  - Added a Warden history scan in Settings to preview eligible failed imports and, after confirmation, add fingerprints to the local list.
  - Added API support for dry-run and confirmed scans, including result and skip counts.
  - NZB metadata now includes poster and publication date when available.

- **Documentation**
  - Documented the history scan workflow, eligibility rules, and dry-run behavior.

- **Tests**
  - Added coverage for scanning, deduplication, skipped entries, concurrent scans, and confirmation flows.


job-bot-ai:
Both findings addressed in 97e0ba7.

**`WardenStore.MarkDeadMany` silently swallowing a failed write** — you're right that this was worse than the single-verdict `MarkDead` case it inherited its handling from. Disposing the transaction rolls the batch back, `LocalCount` is unchanged, and the importer reported `added: 0` — indistinguishable from a genuine no-op. It now returns whether the batch committed, logs at `Warning` instead of `Debug` (a rolled-back batch of thousands isn't the same as one best-effort verdict), and `WardenHistoryImporter` throws rather than reporting a count it can't stand behind.

**The frontend catch block** — also correct, and worse than it looked: a failed write discarded counts that cost a full history scan to produce, then told the user the *scan* had failed. Now only a failed scan closes the modal, and the message names the step that actually failed.

Tests added for both paths: `Import_ThrowsWhenTheWardenWriteFails` drives a real unopenable database rather than a mock (`Initialize` tolerates it, which is exactly the state where a silent failure hides), and a frontend case asserts the modal stays open with the counts intact after a failed write.

Backend 12/12, frontend 4/4.

hoivikaj (commented):
**Short answer: the premise is accurate, but it is not a defect — it's an optional enhancement with a narrow audience and an unverified payoff. You can safely not merge it.**

## What the PR gets right

- `WardenStore.MarkDead` is only called from the Stremio/search-profile paths: `backend/Api/Controllers/Profiles/ProfilePlayController.cs#L442`, `backend/Services/PreflightOrchestrator.cs#L131`, and `backend/Services/Watchtower/WatchtowerService.cs#L928`. Nothing under `backend/Queue/` or `backend/Api/SabControllers/` touches Warden.
- The NZB parser does discard `poster`/`date` (`backend/Models/Nzb/NzbDocument.cs#L98`).
- Failed history rows keep `NzbBlobId`, `FailMessage`, and `TotalSegmentBytes` (`backend/Database/Models/HistoryItem.cs`).

So yes: an *Arr-only install will have 0 fingerprints forever.

## Why that isn't a problem to fix

1. **Warden is never consulted on the *Arr path.** `IsDeadAnywhere` is only checked in `backend/Services/SearchProfileService.cs#L476`, `backend/Api/Controllers/Profiles/ProfilePlayController.cs#L252`, preflight, and Watchtower — all Stremio/search-profile features. The docs describe Warden as "the readiness and safety layer around search and playback" (`docs/features/warden-watchdog-preflight.md`). An empty list on an *Arr-only install is the designed steady state; populating it changes nothing about what Sonarr/Radarr grab. The PR's motivating comment ("the same dead release could be grabbed again and again") isn't fixed by the PR either — `adådfile` still doesn't check Warden.
2. **Re-grab protection already exists on the queue path.** `HealthCheckService.CheckCachedMissingSegmentIds` fails re-grabs of known-missing articles at step 0 (`backend/Queue/QueueItemProcessor.cs#L523`), `ThrowIfRecentlyRejectedAsync` blocks health-rejected releases (`backend/Queue/QueueItemProcessor.cs#L101`), and *Arr itself blocklists anything InfiniDysk reports as `Failed` via SAB history.
3. **The only beneficiary is a hybrid user** (grabs via *Arr *and* uses search profiles/Stremio/Watchtower), and for them it's a pre-seeding convenience, not a correctness fix — their list already grows from real verification failures on that path.

## Efficacy is asserted, not shown

The PR says the fingerprint is "computed exactly as search results are," but the inputs differ: search results hash the indexer's `enclosure length` / `poster` / `usenetdate` attrs (`backend/Clients/Indexers/NewznabClient.cs#L239-L277`), while the importer hashes summed `<segment bytes>` plus the *first* `<file>`'s `poster`/`date`. Any byte difference in size or a midnight-UTC day-bucket boundary yields a fingerprint that never matches. No evidence is offered that even one minted fingerprint matched a live search hit. Also, `"Missing rar volumes detected."` comes from a size-mismatch check (`backend/Queue/FileAggregators/RarAggregator.cs#L294-L299`) that can fire on an incomplete NZB, not only dead articles — and these fingerprints would flow into exports/GitHub backups/shared sources.

## Recommendation

Decline or park it as a feature request rather than treating it as a gap. If you want to keep the door open, the one thing to ask the author for is proof that a fingerprint minted from their history matched a subsequent search-profile result on their indexer — without that, it's ~1,100 lines (new endpoint, UI, contract, `NzbFile` model change) for a list that may never fire.

